PR #10, opened 2016-06-20
Add failing test case for string comparison in if

annismckenzie:
The simple test case I added is failing and I'm wondering if I'm doing something completely outlandish here. Although this particular example is contrived I have a real need in a project of mine for this to work.

Can you take a look, please?


jhsx:
@annismckenzie it's a bug in the expression evaluation, thanks for find out and the test case. i will try to fix this tomorrow.

PR #11, opened 2016-06-22
Add failing test case for integer comparison

annismckenzie:
There's another one – can you take a look? It fails for the fix that you added just a few hours ago.


jhsx:
True, forgot to cast right side values to same kind of the left side. thanks to report, merging the test case.

PR #12, opened 2016-06-23
Add test cases for Go-style map key access

annismckenzie:
I added a few test cases that crash or don't compile yet but I think would be useful to make the template language more feature-complete. These revolve around having `isset` work with non-existing keys and using the two-return variant of accessing a map key. Especially for the latter case I often needed to write custom functions on my structs to check for map keys being defined.

These are just nice-to-haves wrapped in failing test cases and if you don't like the syntax you may also close this pull request – in that case we should consider documenting or discussing on how to handle these cases instead so others don't run into the same trouble I did. :)


jhsx:
Merged manually, into a new branch.

PR #37, opened 2016-09-14
Add ability to use an http.FileSystem for looking up templates

annismckenzie:
Together with a tool like `vfsgen` this can be used to embed the templates into the compiled go binary. This makes deployments simpler and faster but should be used in conjunction with the development mode and a stubbed (read: nil) http.FileSystem.

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"net/http"
 	"reflect"
 	"runtime"
 	"strings"
@@ -353,6 +354,16 @@ func TestFileResolve(t *testing.T) {
 	//for key, _ := range set.templates {
 	//	t.Log(key)
 	//}
+}
+
+func TestFileSystemResolve(t *testing.T) {
+	set := NewHTMLSet()
+	fs := http.Dir("testData/includeIfNotExists")
+	set.SetFileSystem(fs)
+	RunJetTestWithSet(t, set, nil, nil, "existent", "", "Hi, i exist!!")
+	RunJetTestWithSet(t, set, nil, nil, "notExistent", "", "")
+	RunJetTestWithSet(t, set, nil, nil, "ifIncludeIfExits", "", "Hi, i exist!!\n    Was included!!\n\n\n    Was not included!!\n\n")
+	RunJetTestWithSet(t, set, nil, "World", "wcontext", "", "Hi, Buddy!\nHi, World!")
 }
 
 func TestIncludeIfNotExists(t *testing.T) {
--- a/template.go
+++ b/template.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -37,6 +38,7 @@ import (
 // create a set with jet.NewSet(escapeeFn) returns a pointer to the Set
 type Set struct {
 	dirs              []string             // directories for look to template files
+	fsCache           http.FileSystem      // lookup templates using an http file system
 	templates         map[string]*Template // parsed templates
 	escapee           SafeWriter           // escapee to use at runtime
 	globals           VarMap               // global scope for this template set
@@ -90,6 +92,12 @@ func (s *Set) AddPath(path string) {
 	s.dirs = append([]string{path}, s.dirs...)
 }
 
+// SetFileSystem sets the boxed file system for use when looking up templates.
+func (s *Set) SetFileSystem(fs http.FileSystem) *Set {
+	s.fsCache = fs
+	return s
+}
+
 // AddGopathPath add path based on GOPATH env to the lookup list, when loading a template the Set will
 // look into the lookup list for the file matching the provided name.
 func (s *Set) AddGopathPath(path string) {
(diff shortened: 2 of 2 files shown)


annismckenzie:
Don't freak out – this is just a proof of concept. Can you take a look?

jhsx:
I don't know much about tools like vfsgen but i like the idea. I think we can implement this, but we can copy the http.FileSystem interface in that way we will not depend on the http package.

VojtechVitek:
This is pretty cool, @annismckenzie. I'll try it out.

annismckenzie:
@VojtechVitek: getting this to work in concert with the development mode is a bit tricky and you'll need some `go generate` magic and go build tags for that. Holler if you need help (find me [on Gitter](https://gitter.im/CloudyKit/jet)).

jhsx:
@annismckenzie, i been thinking of add a kind of custom loaders, in that way we can have a an interface that check for template existence and get template bytes, in that way we can implement our own loaders.

To implement would be simple, in the template set we walk the list of template loaders and call Exists on each one, the first one to return success we will load the template data from him, if none of they return then, we would load from the filesystem.

We can create sub packages with each loader.

What do you think ?
Per example 
```go


type Loader interface {
	Exists(name string) (bool, error)
	Load(name string) ([]byte, error)
}

// one way to use

func init () {
     set.AddLoader(loaders.NewHttpFileSystem())
}

```

VojtechVitek:
Imho, Sets are just a convenient abstraction over OS files mainly because of lazy loading and cache functionality. Thinking of it a bit more, we don't really need that with in-memory files (like from shurcooL/vfsgen or spf13/afero).

Maybe we don't need another loader or abstraction on top of it, but a simple function like this:

```go
func NewTemplate(r io.Reader) (*Template, error)
```

So people can easily load templates from their os.File or http.File on demand like this:
```
f, err := os.Open("./tpl.jet") // or vfs.Open("/tpl.jet")
tpl, err := NewTemplate(file)
```

VojtechVitek:
hmm, that wouldn't solve dependencies between templates, of course

VojtechVitek:
Actually, I start leaning towards this implementation.

annismckenzie:
I like the `Loader` interface idea very much. How would you propose we proceed, @jhsx?

@VojtechVitek: yes, the template dependencies wouldn't be resolvable this way so I'm a 👎  on that.

VojtechVitek:
Another idea:

```go
type SetFS struct {
    Set
}

// override methods
func (s *Set) loadFromFile(name, fileName string) (template *Template, err error) {}
func (s *Set) fileExists(name string) (string, bool) {}


set := NewHTMLSetFromFS(httpFS) // returns *SetFS
```

annismckenzie:
That will not work, as can be seen here: https://play.golang.org/p/ohw9QxUZje.

In short, both of these private methods are called on `Set`, never on `SetFS `.

VojtechVitek:
of course, it'd require some refactoring

jhsx:
@annismckenzie let's proceed first by adding the loader interface, but i am without time now. So, if someone want to send a PR.

annismckenzie:
I will send one over the coming weekend.

VojtechVitek:
Thanks @annismckenzie!

VojtechVitek:
btw: Wouldn't be
```go
type Loader interface {
	Load(name string) ([]byte, error) // or (io.Reader, err)
}
```
enough? Why do we need `Exists()` method at all? Imho, `Load()` could return different errors and you could check for IsExist() like in os pkg https://golang.org/pkg/os/#IsExist. What do you think?

jhsx:
@VojtechVitek agree,

@annismckenzie as pointed out by @VojtechVitek, we only need the Load method, lets drop the Exists from the Loader interface{}

jhsx:
The question now is should support multiples Loader or only one per Set ?

VojtechVitek:
imho, one loader per set is totally OK

There are things like https://github.com/shurcooL/httpfs union pkg to mount multiple FSs together, just in case

VojtechVitek:
Also, I'd prefer `io.Reader` (or `io.ReadCloser`) as a return type from the `Load()` method. What do you think?

annismckenzie:
No need to introduce a new interface then because `http.FileSystem` looks essentially the same:
```go
type FileSystem interface {
	Open(name string) (File, error)
}
```

Both of the implementations we are using now (reading from a real file system and reading from a virtual one as contained in this PR) and would also make it possible to use https://github.com/shurcooL/httpfs right out of the gate.

VojtechVitek:
@annismckenzie I don't think we want to introduce `http` dependency for Jet.  Let's just use minimal interface, which doesn't require all the `http.File` craft (Seeker, Stat, Readdir).

```go
type Loader interface {
	Open(name string) (io.ReadCloser, error)
}
```

seems like everything what we need - it will work for both os.File and http.File.

annismckenzie:
You're right.

VojtechVitek:
@annismckenzie hey Daniel, sorry to bug you - but are you planning to work on this feature? If not, tell us and we could help you implement/deliver it, since we wanna use Jet+vfsgen for our production email templates within 2 weeks.

annismckenzie:
So sorry for not responding here. Please, go ahead with an implementation. I use the version in this PR for work and it's worked for the past 6 months – just swamped right now. :-/

VojtechVitek:
OK, I'll see if we can take a look later this week or over the weekend and get back to you.

annismckenzie:
Closing this in favor of #48.

PR #42, opened 2016-11-06
Test and fixes for nested interface maps

oliverkofoed:
And also a ParseInline method that does not set development mode.

### DIFF
--- a/template.go
+++ b/template.go
@@ -167,6 +167,17 @@ func (s *Set) resolveNameSibling(name, sibling string) (newName, fileName string
 	return
 }
 
+// ParseInline parses a template given inline
+func (s *Set) ParseInline(name, content string) (*Template, error) {
+	sc := *s
+
+	sc.tmx.RLock()
+	t, err := sc.parse(name, content)
+	sc.tmx.RUnlock()
+
+	return t, err
+}
+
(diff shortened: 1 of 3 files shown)


annismckenzie:
Thanks for raising this!

A couple of things:
1) I cross-linked the setting of the `developmentMode` in `Parse`, let's see if this was really intended or not (meaning that your `ParseInline` should be removed again)
2) In the meantime, can you squash these two commits? The `woops.` commit is kinda bad for the history. :)
3) You're proposing to merge the master branch of your fork, can you create a new branch instead? This way you can continue to pull in changes from the base project.

template.go
```diff
@@ -167,6 +167,17 @@ func (s *Set) resolveNameSibling(name, sibling string) (newName, fileName string
 	return
 }
 
+// ParseInline parses a template given inline
```
jhsx: This functions has same use as Parse and will not work well with import and extends, we set the developmentMode to true to prevent the nested templates to be cached. In next version we should add special flags with clear meaning of what they control, developmentMode is not clear what is used for. @annismckenzie

jhsx:
@oliverkofoed i like to use your tests in this pull request, would you mind to send a pull request with only the tests ?

Thanks

PR #48, opened 2017-02-08
Add Loader interface, Add OS and http.FileSystem loaders

VojtechVitek:
Follow-up from discussion in #37


VojtechVitek:
@jhsx @annismckenzie can you take a look?

I ended up with this:
```
// Loader is a minimal interface required for loading templates.
type Loader interface {
	// Open opens the underlying reader with template content.
	Open(name string) (io.ReadCloser, error)
	// Exists checks for template existance and returns full path.
	Exists(name string) (string, bool)
}
```

annismckenzie:
Wow, this looks very good – I like it a lot! 🎉   🎉   🎉

loaders/httpfs/loader.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"io"
+	"net/http"
+)
+
+type httpFileSystemLoader struct {
+	http.FileSystem
+}
+
+func NewLoader(fs http.FileSystem) httpFileSystemLoader {
```
jhsx: In general is better to not return unexported types in exported functions, better we return the interface instead.
```go 
func NewLoader(fs http.FileSystem) jet.Loader
```

loaders/osfs/loader.go
```diff
@@ -0,0 +1,58 @@
...
+	"path"
+	"path/filepath"
+)
+
+// osFileSystemLoader implements Loader interface using OS file system (os.File).
+type osFileSystemLoader struct {
+	dirs []string
+}
+
+func NewLoader(dirs ...string) *osFileSystemLoader {
```
jhsx: same applies here, for unexported types.

loaders/httpfs/loader.go
VojtechVitek: What's the benefit?
Imho I saw the opposite argument/suggestion from someone in the Go core team. But I might be wrong.

jhsx: I've seen the same arguments as mine in multiple times and also by some of the Go core members.
You can check here https://golang.org/doc/effective_go.html#generality.
Some of the problems with this is that golint will complain and i think that go vet also,  another problem is that go doc will not show the type because is unexported, and also make hard to people to use your type outside your package.

VojtechVitek: Some resources that I googled right now:

```
A great rule of thumb for Go is accept interfaces, return structs.
–Jack Lindamood
```

https://dave.cheney.net/2016/08/20/solid-go-design
http://idiomaticgo.com/post/best-practice/accept-interfaces-return-structs/
https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8#.ervtbgrdp
https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a#.d0zdjc6pm

Pls let me know if I should change the return type or not. I can of course do it anyway, if you want.

VojtechVitek: Nice, https://golang.org/doc/effective_go.html#generality explains it quite well. So it's really about "unexported" structs only, I get it now. Thank you for sharing! 👍

I'll push another commit in a bit.

VojtechVitek: @jhsx Ha, it's not that easy:
```
can't load package: import cycle not allowed
package github.com/CloudyKit/jet
       	imports github.com/CloudyKit/jet/loaders/osfs
       	imports github.com/CloudyKit/jet
import cycle not allowed
package github.com/CloudyKit/jet
       	imports github.com/CloudyKit/jet/loaders/osfs
       	imports github.com/CloudyKit/jet
```

I'd have to put the Loader interface into a separate package, so it could be imported from both jet (template.go) and loaders.

jhsx: Ya, was thinking about this. i prefer that we move the osFileSystemLoader into the jet package since this is the default loader and will be imported any way.

VojtechVitek: sure, makes sense

VojtechVitek: @jhsx 
```
import cycle not allowed in test
package github.com/CloudyKit/jet (test)
       	imports github.com/CloudyKit/jet/loaders/httpfs
       	imports github.com/CloudyKit/jet
```

I'll have to rename the `*_test.go` package and import `jet` from within tests.

jhsx: It should not affect the other loaders, are we importing httpfs in the jet package ?

VojtechVitek: hmm, that's also not possible

`eval_test.go` uses some unexported fields from the `jet` pkg. So this file must remain in `jet` pkg.

And I can't simply move the `httpfs` tests to a different pkg, since it's using functions like `RunJetTestWithSet()`.

VojtechVitek: @jhsx only the test itself imports httpfs. Not the actualy library.

VojtechVitek: @jhsx I moved the osfs pkg into the Jet core, as you suggested. Not sure how to fix the import issue, though. We could move the tests into a sub-directory - or just leave the the `jet.Loader` interface as return value for the time being.

What do you think?

jhsx: I will make a new package jettest that we can later use to move the httpfs tests to his own package. let's return the original type for now then.

So, i will merge you commits now, again thanks for the pull request.

VojtechVitek:
@annismckenzie @jhsx nice, thanks for your help on this one! 👍

PR #50, opened 2017-02-19
Improve custom loaders

annismckenzie:
I got to play around with the new loaders you guys added in #48 and there are a few things that caught my eye while switching from my naive implementation (the one in #37).

Because we're always loading templates from the file system in development and only switch to a binary packaged `http.FileSystem` I needed to copy over your `osFileSystemLoader` implementation – it wasn't exported so I couldn't instantiate it myself. That made me write another custom loader to stitch that and the `httpfs` loader together. All in all I felt like this could be improved so here goes:
1. I made the `osFileSystemLoader` public so it can be used on its own and I didn't see any disadvantage in keeping this private to the library.
2. I added a simple multi loader implementation that is still a `jet.Loader` but can be used with a list of custom loaders.

Now, would you accept this into the core? I don't want to continue with it if you aren't.

If you are, however, then I'd like to polish it with some tests, a nice example of using `vfsgen` in an app that loads templates from the file system in development but switches to a packaged `http.FileSystem` when compiled and also add a bit of documentation of how to use this to the wiki.

### DIFF
--- a/loader.go
+++ b/loader.go
@@ -40,19 +40,19 @@ type hasAddGopathPath interface {
 	AddGopathPath(path string)
 }
 
-// osFileSystemLoader implements Loader interface using OS file system (os.File).
-type osFileSystemLoader struct {
+// OSFileSystemLoader implements Loader interface using OS file system (os.File).
+type OSFileSystemLoader struct {
 	dirs []string
 }
 
 // Open opens a file from OS file system.
-func (l *osFileSystemLoader) Open(name string) (io.ReadCloser, error) {
+func (l *OSFileSystemLoader) Open(name string) (io.ReadCloser, error) {
 	return os.Open(name)
 }
 
 // Exists checks if the template name exists by walking the list of template paths
 // returns string with the full path of the template and bool true if the template file was found
-func (l *osFileSystemLoader) Exists(name string) (string, bool) {
+func (l *OSFileSystemLoader) Exists(name string) (string, bool) {
 	for i := 0; i < len(l.dirs); i++ {
 		fileName := path.Join(l.dirs[i], name)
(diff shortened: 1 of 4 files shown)


annismckenzie:
Maybe @VojtechVitek can also take a look because I did touch the code he contributed in #48.

annismckenzie:
Nice, thank you! I'll finish this up tomorrow.

VojtechVitek:
SGTM too.

@jhsx Now question is - do we want to expose `OsFileSystemLoader` as public struct, or do we want `loaders/osfs` subpkg to return `jet.Loader`?

```go
import "github.com/CloudyKit/jet/loaders/osfs"

var fs = osfs.New() // jet.Loader
```

annismckenzie:
For the future: I'll only open a pull request when I'm done. I wasn't really and I just wanted your opinion – I'll open a separate pull request with the polishing.

jhsx:
@VojtechVitek since this is the default loader i think is better to maintain in the jet package.
@annismckenzie one thing we missed was a constructor, in order to make easy to use we need to add a NewOSFileSystemLoader func, per example: 
```go
var ViewSet = jet.NewSetLoader(nil, multi.NewLoader(
	httpfs.NewLoader(http.Dir("./embed-views")),
	jet.NewOSFileSystemLoader("./views"),
))
```

annismckenzie:
Yes, you're right – do you want me to add it?

@VojtechVitek: no, we can't have it return `jet.Loader` as you'd lose the ability to call `AddPath()` on that (since those are not part of the interface).

jhsx:
I was finish a message when i saw your message. So, my bad, next time i will ask if is ready to merge. Ya.
Thanks for contributing 😄

PR #67, opened 2017-10-12
Raising a panic in the includeIfExists helper when template exists but fails to render.

diogogmt:
In our template framework which is powered by Jet we always use the helper `includeIfExists` to load templates from dynamic variables which we don't know what the value will be before hand and thus can't rely on the fact that the template they reference will exist.

However, we are experiencing different behavior on how rendering errors are handled by the helper.

1 . When running Jet in development mode and loading the templates from the disk, the template rendered by `includeIfExists` errors out and fails silently.
2. When running Jet in production mode and loading the templates via a virtual file system the `includeIfExists` helper panic's but the error is omitted.

This PR raises a panic in the `includeIfExists` helper when the template exists but fails to render.
The error is then properly bubbled up the chain when loading from disk or a VFS.

It would be great if we could suppress the the first panic and only panic in the helper since the panic by itself doesn't have a helpful message.


annismckenzie:
Would you be able to add test case for this? I was just about to introduce `includeIfExists` in some of our templates as well…

annismckenzie:
Forgot to say thank you for bringing this up so polished and with code to boot!

annismckenzie:
Thanks for the test cases. Are you satisfied with the changes? I would need to do a rebase and change a few things before we can merge this though:
1. The .travis.yml file update is a separate issue/MR (#69) – can you undo that change?
2. The last commit is just a fixup. Squash it into the previous commit.

Do you feel comfortable around git rebase or do you want me to take care of it?

diogogmt:
@annismckenzie OK, I'll update the PR later this week.

annismckenzie:
So sorry, didn't see you did the rebase already. Thaaaaank you! 🎉 

@jhsx I feel comfortable merging this now. No others tests were harmed in the making of these changes. 😅

Issue #72, opened 2017-11-13
Reloading templates in production


japettyjohn:
I have a website using jet and it looks like unless I'm in development mode I won't have a way updating existing pages. For something like a CMS (which is what I'm working with) this is an issue.

My plan right now is to add an UnloadTemplate call which removes the old one from the cache, then the next hit on the template get reloaded as per usual. It could also be useful in the case 

Is that the right approach?

japettyjohn:
[This is what I'm looking at](https://github.com/japettyjohn/jet/commit/a52168e2d1911f309eda148340b4807c9285ede2), will do a PR when I've done a bit more testing.

japettyjohn:
This is working now and I've submitted it on #73.

jhsx:
@japettyjohn a simple way to achieve this is to just create an new set. This is not the most optimized way, but i think still fine cause i think template reloading will not happen so often.

japettyjohn:
@jhsx Agreed. 

In the meantime I reviewed what needs to reload and identified it did not need any templating inside of it, rather it was static body content which was wrapped in a template. 

As such I made a template which takes the content as a buffer and writes it out. Forgoing the template reload and avoids storing every page in memory as a "template".

annismckenzie:
I agree with @jhsx on this one. If you really need to be able to reload templates, you should wrap the set and guard it with a mutex. While that could cause contention it's probably the least risky way and shouldn't impact performance that much if the render/reload ratio is really small. You'll probably want to separate your normal templates and the ones that need hot reloading by having two different template directories. The edge cases to support this in the core are so numerous that I really don't want us to do this.

Hope you're not mad but I'm closing this as a wontfix.

PR #73, opened 2017-11-14
added template unload

japettyjohn:
Template reloading is all or nothing with development mode as the only reload option. Hot loading is supported so this adds the missing unload for that scenario.

### DIFF
--- a/template.go
+++ b/template.go
@@ -314,6 +314,17 @@ func (s *Set) LoadTemplate(name, content string) (template *Template, err error)
 	return
 }
 
+func (s *Set) UnloadTemplate(name string) {
+	if s.developmentMode {
+		return
+	}
+
+	s.tmx.Lock()
+	defer s.tmx.Unlock()
+	delete(s.templates, name)
+	return
+}
+
 func (t *Template) String() (template string) {
 	if t.extends != nil {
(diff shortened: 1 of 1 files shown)


annismckenzie:
The code looks easy enough, thanks for that.

Can you add:
1. A comment according to go convention
2. A unit test

I know that the comments in these parts of the code base are sparse but I'd like to start somewhere.

jhsx:
Firstly we need to add an way to mark the template as unloaded, we can set Template.set=nil and check in the evaluation if the imported or extended template has .set setted to nil, then try to reaload and update the pointers.

jhsx:
And thanks for contributing :smile:

japettyjohn:
@jhsx AST complications could be a nest of troubles, avoidable I think given I'm currently only trying to reload, not generally unload. See my latest commits for that.

@annismckenzie I've added the comments, test is outlined but not done.

jhsx:
You will need to update blocks map for every template, that imports or extends the template being reloaded.

Please check, https://github.com/CloudyKit/jet/blob/master/parse.go#L165 and https://github.com/CloudyKit/jet/blob/master/eval.go#L485

annismckenzie:
@japettyjohn see my comment on #72. Can you close this, please?

PR #78, opened 2018-03-31
Support platform-appropriate path-separators

9999years:
Changes `"path"` to `"path/filepath"` wherever it's imported (in `loader.go` and `template.go`) to support platform-dependent path-separators. Previously, Windows-style paths would cause template load errors on relative imports because `"/"` was hardcoded as the path separator in `template.go`:L147, causing `include` and `extends` statements to not correctly find sibling-files.

Also improves the error messages in `eval_test.go`.

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -108,7 +108,7 @@ func RunJetTestWithSet(t *testing.T, set *Set, variables VarMap, context interfa
 	}
 
 	if err != nil {
-		t.Errorf("Parsing error: %s %s %s", err.Error(), testName, testContent)
+		t.Errorf("Parsing error: `%s` in test `%s` of content `%s`", err.Error(), testName, testContent)
 		return
 	}
 	RunJetTestWithTemplate(t, tt, variables, context, testExpected)
--- a/loader.go
+++ b/loader.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"io"
 	"os"
-	"path"
 	"path/filepath"
 )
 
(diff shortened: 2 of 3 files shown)


annismckenzie:
Thanks for your contribution! I'll see about getting this merged next week – I want to setup AppVeyor testing first (running on Windows machines). Just to clarify: with these changes the test suite is green on your machine, right?

9999years:
Test suite is green but I didn’t add any tests to work specifically with backslash-delimited paths.

annismckenzie:
Wait, but in your templates you wouldn't want to use backslash-delimited paths when using `{{ include }}` or `{{ import "something/in/a/subfolder.jet" }}` because then that project wouldn't work on macOS or Linux, right? 🤔

annismckenzie:
I'm so sorry that this took so long. Sigh.

So, I finally added AppVeyor to run our test suite on Windows (not merged yet, see #88). Unfortunately, and as I suspected with my previous comment, your changes break some tests (see https://ci.appveyor.com/project/annismckenzie/jet/build/20 at the end).

Please also see https://ci.appveyor.com/project/annismckenzie/jet/build/17 which is the latest build of #88 and actually runs the asset packaging example on Windows both in the compiled version (with assets inside the binary and using the HTTP loader) as well as using `go run main.go` using the assets/views from the OS file system. I added a bunch of Jet template features to that example so that you can see that it works.
I still firmly believe that you're using backslashes in your Jet templates for including or importing other templates – like `{{ import "something\in\a\subfolder.jet" }}`. Don't do that and it'll work fine. Or am I missing something? If you can add a failing test case to this branch I can push it over to my fork and run it in AppVeyor.

I'm gonna look at #79 next.

annismckenzie:
Closing this as it actually breaks the HTTP loader. Always use forward slashes for including, importing and extending – it works on Windows, Linux and macOS (the latter two I'm constantly testing with our production app and Windows will get CI via #88).

Anyone who has a failing test case please reopen – or better yet, open a PR after #88 is merged. :)

PR #79, opened 2018-03-31
Added a jet command-line app

9999years:
Adds a simple command-line app in `cmd/jet` to render templates; I think this will allow people to get started writing and executing templates more easily.

Is it a good idea / something you’d like to have in the repo? If it makes more sense to you, I’d be happy keeping it separate.

Build is just `go build` or `go install`. There’s about a dozen tests too (`go test`).

Usage:

```
Usage: jet [options] [-template] TEMPLATE_NAME
  -dir string
        The directory to search for templates in (default "./")
  -template string
        The filename of the template to render
NOTE: TEMPLATE_NAME can be given either as a named flag or positionally as
the last argument
```


9999years:
Ugh, getting an error because of (I think?) a difference in the `flag` API in the Go 1.8/1.9 on Travis and the Go 1.10 I’m developing on. Any clue where I can read old Godocs to make this compatible?

9999years:
OK, found some old docs [here](https://golangdoc.github.io/pkg/1.8/flag/index.html) and fixed it. All green now.

annismckenzie:
The problem I see with this is that as soon as you have a template that needs variables or the context, i.e. pretty much why you'd wanna use a template engine, the rendering will fail because you can't execute the template without those and there's no way to pass via command line flags. So the use case is pretty limited. What I could see is having a cli tool that checks your templates for parse errors, i.e. a Jet template linter. That could also be used in CI – point it at the folder your templates are in and it'll fail if there are syntax errors. What do you think?

PR #88, opened 2018-07-01
Run test suite on Windows via AppVeyor

annismckenzie:
I finally got around to adding AppVeyor for running the test suite on Windows in our PRs. It's especially related to #78. The good news is that Jet already works correctly on Windows. :+1: AppVeyor is free for open source projects.

I marked it as a work-in-progress because there's a few issues we still need to discuss:
- [x] Windows normally have Git's `core.autocrlf` set to auto which breaks our tests because the template comparisons are done verbatim and Windows has CRLF line endings, while the rendered templates in memory only have LF. I worked around this using a crude string replace (see code, ugh). The other way would be to set `core.autocrlf` to `input` and then it'll work.
- [x] As an addition I enhanced the asset packaging example to also be executed in AppVeyor to test a real app (both using the compiled version as well as just using `go run`) – thought that'd be good.
- [x] AppVeyor needs to be enabled in the CloudyKit organization.
- [x] Adding additional Go versions (the last two) via the matrix feature of AppVeyor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Jet Template Engine for Go [![Build Status](https://travis-ci.org/CloudyKit/jet.svg?branch=master)](https://travis-ci.org/CloudyKit/jet) [![Build status](https://ci.appveyor.com/api/projects/status/5g4whw3c6518vvku?svg=true)](https://ci.appveyor.com/project/CloudyKit/jet)
 
-[![Join the chat at https://gitter.im/CloudyKit/jet](https://badges.gitter.im/CloudyKit/jet.svg)](https://gitter.im/CloudyKit/jet?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Join the chat at https://gitter.im/CloudyKit/jet](https://badges.gitter.im/CloudyKit/jet.svg)](https://gitter.im/CloudyKit/jet)
 
 Jet is a template engine developed to be easy to use, powerful, dynamic, yet secure and very fast.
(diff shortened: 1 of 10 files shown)


annismckenzie:
Alright, this is now done and also runs the test suite on Windows using Go 1.9 and 1.10.

annismckenzie:
@jhsx I still need you to add AppVeyor as an app to the Jet repo or approve the request I did from my fork or go to https://github.com/marketplace/appveyor and add it to the CloudyKit organization (it's free for open source projects and only has access to public repositories). Please? 😬

jhsx:
@annismckenzie sorry for the time that i taken to answer, only saw your email yesterday. I gave you the admin access, go ahead.

Thanks.

annismckenzie:
Wow, GitHub's deprecation of their Services and the replacement with GitHub Apps is a real pain.

annismckenzie:
@jhsx Somehow, this is all very bad. You'll have to add AppVeyor with public access via the marketplace (URL in my last comment) to the CloudyKit organization because just using AppVeyor via their webhook doesn't work – what a pain. So sorry about all of this but I really want to get this done. :/

jhsx:
@annismckenzie i just added appVeyor to the organization and project can you check ?

jhsx:
https://ci.appveyor.com/api/projects/status/5g4whw3c6518vvku?svg=true

jhsx:
[Build status](https://ci.appveyor.com/project/CloudyKit/jet)

PR #89, opened 2018-07-02
Implement multiline handling for yield/block/function definitions

### LINKED ISSUE #86: yield/block/function invocation statements cannot be multiline
If there is newline break inside a yield statement e.g after a , the parser dies with "unclosed action"

annismckenzie:
This fixes #86 but isn't just for `yield`s but because the parser easily handles this as "inside-actions" it's now also possible when invoking functions and defining blocks.

As a quick example these would have failed with `unclosed action` before but work now:

```
{{block col(
	columns,
)}}
	…

{{yield row(
	cols="12",
	rowClass="highlight-row",
) content}}

{{replace("My Name Is II",
	" ",
	"_",
	-1,
)}}
```

Edit: the edge case below is fixed.

I noticed one edge case:
```
{{replace("My Name Is II",
	" ",
	"_",
	-1,
)}}
```

~Here the dangling comma after the -1 trips up the parser and it fails with `unexpected token ")" on operand` because it expects the decimal number to continue. One way to fix it is to disallow dangling commas (which are actually a side effect of the lexer) but that syntax feels very Go-like. This edge case fails when parsing the template so it's not a runtime failure upon template execution so I'd like to keep it if we can live with it.~

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -243,6 +243,37 @@ func TestEvalBlockYieldIncludeNode(t *testing.T) {
 		`{{import "BlockContentLib"}}{{yield row(cols="12") content}}{{cols}}{{end}}`,
 		"\n    <div class=\"row\">\n        \n            \n    <div class=\"col 12\">12</div>\n\n        \n    </div>\n")
 
+	JetTestingSet.LoadTemplate("BlockContentLib2", `
+		{{block col(
+			columns,
+		)}}
+			<div class="col {{columns}}">{{yield content}}</div>
+		{{end}}
+		{{block row(
+			cols="",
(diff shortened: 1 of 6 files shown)


mvrhov:
As for the dangling comma. I've looked a bit at the parser about a week ago with the debugger, but I hadn't had enough time to explore deeply .. 
I remember vaguely that there are peek and "backtrace" functions. So if after `,` is unlimited number of spaces and then `)`...

annismckenzie:
I fixed that edge case and the test suite seems to agree. :dancer:

annismckenzie:
Nice, merging!

Issue #84, opened 2018-05-10
Pipelining and yielding


mvrhov:
i'd like to use pipelinig inside a yield expression to pass a translation to a block

`{{yield inputField(id="email", name="email", type="email", label="login.form.label.email"|T:"Email",map(),"account", icon="fa-envelope", required=true)}}`

annismckenzie:
The things we come up with as humans writing template files. 😆

So yeah, the parser is probably failing on the first version. The second one, though, looks correct (I use blocks and yields a lot in our projects). What's the error? Does it fail on the first line or the second? Does the label variable contain the translated content? Also, you may want to try the plain old function invocation syntax which should work with inside the yield as `T("Email", etc.)`.

As for the first version and making it work: I don't think I want to (sorry about that). 🙈 As a member on a team having to grasp that statement… it doesn't look very readable. Add to that, that as you've written it the parser couldn't easily distinguish between when the pipeline ends and the next parameter begins (yes, the next parameter is a `parameter=` token but still).

Just to be clear on who I am: I try to help people out with syntax and questions but haven't actually done any work on the engine (parser, lexer etc.) and I always defer to José. Sorry about that.

mvrhov:
I don't know the error from the top of the head but in both cases it chokes on the |

annismckenzie:
I stand by my comment. Having written a lot of templates I really really don't want to make the first version work. There's just too many things going on. With #89 you will also be able to break up the yield into multiple lines and put `T("login.form.label.email", "Email", …)` as the value of the `label` parameter, making it readable and explicit. Does that sound fair? Please close the issue in that case.

mvrhov:
Even though that you don't like the translate version example... there is the same issue if one want's to use "default" in pipelinig fashion.

e.g `
{{yield inputField(id="email", name="email", type="email", label=label, icon="fa-envelope", required=true error=formErrosMap["email"]|default("") )}}`

I'll say it here again like in another issue. This is the inconsistency with pipelinig that should be fixed.

annismckenzie:
I see your point. I'll put it on the list. The problem that I see there is that it'll probably blow up even if I make the command understand pipelining.

Just out of curiosity, is this just the example or wouldn't you be able to use Go's default return value of an empty string when accessing non-existant map keys? 🤔

[...]

Issue #32, opened 2016-08-26
Iconsistency with add operator


jhsx:
Taking in example

```go
	RunJetTest(t, data, nil, "actionNode_AddIntString", `{{ 2+"1" }}`, "3")
	RunJetTest(t, data, nil, "actionNode_AddStringInt", `{{ "1"+2 }}`, "12")
```

The first test the left operand is a number and the right is a string then the right operand is converted to a number and the expression is evaluated as add expression.

The second test the left operand is a string and the right is a number the expression is evaluated as a string concatenation.

I propose a breaking change in the template language, adding a operator for string concatenation and let numeric operators to work only with numbers.
Like php $val.$val2.

@annismckenzie

annismckenzie:
Yes, I concur with this. That means bumping the version to 3.0.

annismckenzie:
#93 fixes this another way by treating both as arithmetic expressions:
```go
RunJetTest(t, data, nil, "actionNode_AddIntString", `{{ 2+"1" }}`, "3")
RunJetTest(t, data, nil, "actionNode_AddStringInt", `{{ "1"+2 }}`, "3")
```

I'm not sure if that's a good idea though (I'll raise the issue over there too) because isn't the use case of a template engine to allow string concatenation without caring about the individual pieces' type?

As an example:
```
{{ someVariable := 10 }}
{{ prefix := "The answer is " }}
{{ suffix := ". Some more text." }}
{{ prefix + someVariable + suffix }}
```

I'd really expect this to print `The answer is 10. Some more text.`. /cc @maxime1907

This is also related to #83.

mvrhov:
FYI: the twig/jinja etc use ~ as a concat operator. I don't know if this is reserved here, but It's more readable than the dot as dot can easily be lost in the visual clutter.

So if there will be a different concat oeprator, then the examples above should succedd,
however `RunJetTest(t, data, nil, "actionNode_AddIntString", `{{ 2+"1z" }}`, "3")` should fail!

maxime1907:
@mvrhov I agree with adding a special operator for specific use case like "2" ~ "2" output: "22", but ~ operator is mostly used for regex and binary operations ^^
Regarding the changes on the additive function, i made it works that way (same as perl template engine):
- {{ "2" + "2" }} will output: 4
- {{ "2" + "2hey" }} will output: 22hey
- {{ "2" + 2 }} will output: 4
- {{ someVariable := 10 }}
{{ prefix := "The answer is " }}
[...]

PR #91, opened 2018-07-03
Add ability to use custom delimiters

### LINKED ISSUE #68: Add custom delimiters
[https://github.com/golang/go/blob/master/src/html/template/template.go#L351](https://github.com/golang/go/blob/master/src/html/template/template.go#L351)

`func (t *Template) Delims(left, right string) *Template`
Can it be realized?

### LINKED ISSUE #98: Is there currently a way to escape the default delimiters?
I'm aware that custom delimiters are in the works, personally I prefer using the golang conventions of {{ }}, however I am also using Vue which uses {{ }} for its variables.

Is there currently an easy way to escape the {{ }} characters? I tried with \{{ but that didn't seem to work, at the moment I am having to use {{ "{{ vueVariable }}" | unsafe }}

annismckenzie:
This fixes #68 and introduces the ability to use custom delimiters.

Use like this:
```go
var View = jet.NewHTMLSet("./views"))
View.Delims("[[", "]]")
```

```html
<!-- file: "views/home.jet" -->
[[ extends "layouts/application.jet" ]]
[[ block body() ]]
  <main>
    This content will be yielded in the layout above.
  </main>
[[ end ]]
```

TODO:
- [x] test extensively
- [x] add documentation for it to the wiki

This fixes #68 and closes #98.

### DIFF
--- a/lex.go
+++ b/lex.go
@@ -235,7 +235,7 @@ func (l *lexer) drain() {
 }
 
 // lex creates a new scanner for the input string.
-func lex(name, input string) *lexer {
+func lex(name, input string, run bool) *lexer {
 	l := &lexer{
 		name:       name,
 		input:      input,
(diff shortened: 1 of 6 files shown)


annismckenzie:
For now, I'd stay away from a) making comment delimiters dependent upon the delimiter that's used and b) from bumping to v3 unnecessarily. The use case is to use Jet templates on the server in convert with Mustache-style template languages on the client side. Vanilla Mustache uses `{{! }}` for the comments so I think we're fine with `{* *}` which personally I haven't come across anywhere.

If and when we go ahead with #93, #84 and #32 we can talk about it again because for these I really do want to bump the major.

annismckenzie:
I think this has reasonable coverage and will not break anything. I'll rework the Wiki slightly (it only mentions the `Delims` via this PR) and then go ahead and merge it.

annismckenzie:
I updated the wiki: https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax#delimiters

Issue #60, opened 2017-05-05
Add contributors section in the readme


jhsx:

annismckenzie:
I'll take care of this today.

sauerbraten:
How detailed should this be? I would put @jhsx, @annismckenzie and myself, based on the number of commits as well as code additions and deletions: https://github.com/CloudyKit/jet/graphs/contributors

sauerbraten:
Probably closed in 2ada4091acbbc7981ee93f5a91c33fdf539ee4c4.

tooolbox:
Ya, I think we're good to close.

PR #93, opened 2018-07-04
Fix inconsistency with add operator, Added switch/case, Added default block values, Filter block with format function

maxime1907:
Add:
"2" + 1 - output: 3
1 + "2" - output: 3
"2" + "2" - output: 4
"2" + "2z" - output: 22z

Switch/Case with empty case as default :
{{ switch power }}
    {{ case 10 }}
        Value is 10
    {{ end }}
    {{ case }}
        This is default case
    {{ end }}
{{ end }}

Default block:
- Example 1:
{{ Size = 10 }}
{{ default }}
    {{ Size = 20 }}
{{ end }}
Size is {{ Size }} cm
output : Size is 10 cm

- Example 2:
{{ default }}
    {{ Size = 20 }}
{{ end }}
Size is {{ Size }} cm
output : Size is 20 cm

Filter block with format function:
{{ filter format("<-- %v -->") }}
    This is
    formatted text
{{ end }}
output:
<-- This is -->
<-- formatted text -->

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Jet Template Engine for Go [![Build Status](https://travis-ci.org/CloudyKit/jet.svg?branch=master)](https://travis-ci.org/CloudyKit/jet) [![Build status](https://ci.appveyor.com/api/projects/status/5g4whw3c6518vvku?svg=true)](https://ci.appveyor.com/project/CloudyKit/jet)
 
-[![Join the chat at https://gitter.im/CloudyKit/jet](https://badges.gitter.im/CloudyKit/jet.svg)](https://gitter.im/CloudyKit/jet?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Join the chat at https://gitter.im/CloudyKit/jet](https://badges.gitter.im/CloudyKit/jet.svg)](https://gitter.im/CloudyKit/jet)
 
 Jet is a template engine developed to be easy to use, powerful, dynamic, yet secure and very fast.
 


annismckenzie:
What is happening all of a sudden? Ho-ly! This is major. You seem to have a good grasp on the whole parser and lexer business (\*makes-a-mental-note\*).

This is a lot to unpack and we'll need multiple rounds of reviews – hope you're up for that. I might also opt for splitting this up into multiple PRs.

I'll do an initial review this evening.

Also, this needs some emojis: 🎉🎉🎉

template.go
```diff
@@ -371,6 +372,9 @@ func (scope VarMap) SetWriter(name string, v SafeWriter) VarMap {
 
 // Execute executes the template in the w Writer
 func (t *Template) Execute(w io.Writer, variables VarMap, data interface{}) error {
+	if w == nil {
```
jhsx: This default we should skip, we need to fail when the user pass an nil writer. We can even panic with a descriptive error message.

eval.go
```diff
@@ -549,6 +760,38 @@ var (
 	valueBoolFALSE = reflect.ValueOf(false)
 )
 
+func ParseIndexExpr(baseExpression reflect.Value, indexExpression reflect.Value, indexType reflect.Type) (reflect.Value, error) {
```
jhsx: Can you refactor this function in to private ? I think is better to export only the names meant to be used by the user.

annismckenzie:
I'm gonna pull #32 into the discussion:
> I propose a breaking change in the template language, adding a operator for string concatenation and let numeric operators to work only with numbers.
Like php $val.$val2.

I agreed with it then and I agree with it now. Addition should only apply to numeric parts and we should add the dot operator (or something else) to concatenate strings where numeric components are cast to strings. Whatever we decide we should move those code changes to a separate PR that references #32 and if we decide that we instead want it this way then that's fine too. Can you move the commits over to a separate branch, please?

examples/todos/main.go
```diff
@@ -130,4 +130,4 @@ func main() {
 
 	log.Println("Serving on " + port)
 	http.ListenAndServe(port, nil)
-}
+}
```
annismckenzie: Please don't drop empty lines at the end of files, it trips up Git's diff algorithm from time to time and makes staging single-line hunks unreliable. If you don't have it yet, add https://editorconfig.org to your editor and configure it with
```
[*]
end_of_line = lf
insert_final_newline = true
```
for the Jet repository (we can add that config file to the .gitignore if need be).

lex.go
```diff
@@ -100,29 +100,39 @@ const (
...
-	"range": itemRange,
-	"nil":   itemNil,
-	"and":   itemAnd,
-	"or":    itemOr,
-	"not":   itemNot,
-
-	"content": itemContent,
-	"msg":     itemMSG,
-	"trans":   itemTrans,
+	"extends": 	itemExtends,
```
annismckenzie: It doesn't look like you're using go fmt? Always use go fmt in Go projects.

eval.go
```diff
@@ -391,22 +553,71 @@ func (st *Runtime) executeList(list *ListNode) {
...
 						}
+					} else {
+						tmp := []byte(fmt.Sprintf("%v", v.Interface()))
+						optionText.SetText(tmp)
 					}
 				}
 			}
+		case NodeDefault:
+			node := node.(*DefaultNode)
+
```
annismckenzie: Can you drop the surrounding empty lines here and also in the following case statements?

eval.go
annismckenzie: I'd also add it to the Runtime struct like the other expression parsing methods.

annismckenzie:
A couple more things and some comments to sum up how I'd like to continue this:
1. What's the use case {{ filter format("something") }} being added to the language? I can't think of a situation right now where I'd use it. 🤔
2. The switch/case is valuable, as is the default block so these can go in together. :+1:
3. The add operator inconsistency check code should be moved to a separate branch. That will likely also result in this being included in the v2.2.0 release instead of being moved to v3.0.
4. Various nits with code formatting.
5. Test coverage of these changes is 0 which we'll need to change before merging.

In see you committed this straight to your fork's master branch which is a weird thing to do. 😬 So, I can help with all of those chores and nits if you want. You enabled maintainers being able to add commits so just say the word and I'll jump in. It's really great to get contributions and you'll be added to the contributors section in the README which I still haven't done (#60) but will soon. 🎉

mvrhov:
IMO the default shouldn't be a block, but a pipeline "filter".

mvrhov:
No really I don't see the value of a default block.
Who doesn't this work work you? `Size is {{ Size|default(20) }} cm`, because `Size` might have not be defined? The default pipe should handle that.

maxime1907:
I don't really have a preference, i just looked at http://www.template-toolkit.org/docs/manual/Directives.html#section_DEFAULT and thought it may be a good idea but if you have a preference then let's go do this :smile:

mvrhov:
:+1 for default implementation

maxime1907:
Have i done what you wanted me to fix ? If so, can you consider accepting my pull request ? :smile:

annismckenzie:
Thanks for all the changes! I'll be going through this tomorrow. I'm also going to have to do a rebase against the current master.

vendor/github.com/CloudyKit/fastprinter/printers.go
```diff
@@ -178,7 +187,7 @@ func PrintValue(w io.Writer, v reflect.Value) (int, error) {
 		return PrintUint(w, v.Uint())
 	}
 
-	if k == reflect.Float64 || k == reflect.Float32 {
+	if k == reflect.Float64 || k == reflect.Float64 {
```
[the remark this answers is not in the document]
maxime1907: no i just cloned https://github.com/CloudyKit/fastprinter/ and https://github.com/CloudyKit/fastprinter/blob/master/printers.go does contain this error ^^

tooolbox:
@annismckenzie Another use case for "filter" would be to pipe a block through [quotedprintable](https://godoc.org/mime/quotedprintable) when you're rendering emails to send via smtp.

Exciting!  Well done @maxime1907

vendor/github.com/CloudyKit/fastprinter/printers.go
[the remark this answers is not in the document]
annismckenzie: I'll open a separate MR about this later. For now, I reverted to the previous version of the vendored code and opened #94.

annismckenzie:
Oh GitHub! That »collaborators can just push to that branch« is really really weird and I made a mistake that zeroed out your branch. And because I did that it closed the MR and also removed the permission for me to fix it. I did rework your commits and they still contain your author name and email. Can I trouble you to push your commits again and reopen the MR? So sorry about this. 😖

PR #96, opened 2018-07-18
Fix evaluating pointers

annismckenzie:
For now I just added 2 failing test cases. Maybe this isn't just about string pointers but others like int pointers as well.


annismckenzie:
Very nice, merging! Thanks @sarge for reaching out via Gitter and pushing the fix!

PR #97, opened 2018-07-18
Add support for dereferencing pointers. Nil prints as empty string

sarge:
Try this as a possible implementation.

Interesting potential pickle where you could have a chain of pointers, I didn't construct a circular chain but I suppose it is possible.

Not sure about additional allocations.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -16,12 +16,13 @@ package jet
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


annismckenzie:
Nice, I'll pull it into mine for further discussion!

annismckenzie:
A follow-up is at #100.

PR #100, opened 2018-08-03
Add support for pointer fields on structs

c2h5oh:
+ minor cleanup on the last commit for pointer fields

PR #102, opened 2018-08-09
WIP: Fix different behaviour when accessing maps

### LINKED ISSUE #101: Unexpected behaviour in similar templates
I'm passing ```{"a":null}``` (decoded to map[string]interface{}) to template
```
{{ .a }}
```
and receiving error `there is no field or method "a" in map[string]interface {}`
When I change template to 
```
{{ .["a"] }}
```
and everything works (as expected).
But, if I pass ```{"a":{"b":null}}``` to templates 1)`{{.a.b}}` and 2)`{{.a["b"]}}` I'm receiving error 1)`there is no field or method "b" in map[string]interface {}` and 2) nothing (everything ok).

As I understand, for data structure `map[string]map[string]map[string]map[string]map[string]interface{}` template 
```
{{ .a.b.c.d.e.f }}
```
would raise error, but for
```
{{ .a.b.c.d.e["f"] }}
```
everything works fine.

So my question is:
Why jet can't select the last element in chain of map[string]interface{} by using dots (not square brackets)?

And can I push pr that fix this?

vetcher:
This pr fixes Jet Runtime error when accessing nil element of the `map[string]interface{}` via dot.
Should fix #101

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -341,7 +341,7 @@ func TestEvalIndexExpression(t *testing.T) {
 	RunJetTest(t, nil, map[string]interface{}{"nested": map[string]string{"name": "value"}}, "IndexExpressionMap_6", `{{.["nested"].name}}`, "value")
 
 	vars := make(VarMap)
-	vars.Set("nested", map[string]interface{}{"key": "nested", "nested": map[string]interface{}{"nested": map[string]interface{}{"nested": map[string]interface{}{"name": "value", "strings": []string{"hello"}, "arr": []interface{}{"hello"}}}}})
+	vars.Set("nested", map[string]interface{}{"key": "nested", "nested": map[string]interface{}{"nested": map[string]interface{}{"nested": map[string]interface{}{"name": "value", "strings": []string{"hello"}, "arr": []interface{}{"hello"}, "nil": nil}}}})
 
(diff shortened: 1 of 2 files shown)


eval_test.go
```diff
@@ -355,6 +355,8 @@ func TestEvalIndexExpression(t *testing.T) {
 	RunJetTest(t, vars, nil, "IndexExpressionMap_13", `{{nested.nested.nested["nested"]["arr"][0]}}`, "hello")
 	RunJetTest(t, vars, nil, "IndexExpressionMap_14", `{{nested["nested"].nested["nested"].name}}`, "value")
 	RunJetTest(t, vars, nil, "IndexExpressionMap_15", `{{nested["nested"]["nested"].nested.name}}`, "value")
+	RunJetTest(t, vars, nil, "IndexExpressionMap_16_1", `{{nested.nested.nested.nested.nil}}`, "<nil>")
```
annismckenzie: @c2h5oh: wouldn't we also want this to be an empty string instead of the `<nil>` (same as for structs, see #97 and #100)?

annismckenzie:
Awesome, thanks! Merging.

eval_test.go
c2h5oh: Yeah, it's better to be consistent

annismckenzie:
It would be awesome if you were to take a stab at that. 😅

PR #103, opened 2018-08-10
Nil value of map prints empty string

vetcher:
Make accessing to map's nil value prints empty string instead of `<nil>`, similar to #100 and #97 
Fix previous tests and add 2 new.

There is one side effect: when trying to do 
```
{{ .A.B.C }}
```
when A is `map[string]interface{}` and B is `nil` returns error `there is no field or method C in string`

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -36,6 +36,7 @@ var (
 			return &Runtime{scope: &scope{}, escapeeWriter: new(escapeeWriter)}
 		},
 	}
+	emptyString = reflect.ValueOf("")
 )
 
 // Renderer any resulting value from an expression in an action that implements this
@@ -602,7 +603,11 @@ func (st *Runtime) evalPrimaryExpressionGroup(node Expression) reflect.Value {
 					node.errorf("%s is not assignable|convertible to map key %s", indexType.String(), key.String())
 				}
 			}
-			return baseExpression.MapIndex(indexExpression)
+			baseExpression = baseExpression.MapIndex(indexExpression)
+			if baseExpression.Kind() == reflect.Interface && baseExpression.IsNil() {
+				return emptyString
+			}
+			return baseExpression
 		case reflect.Array, reflect.String, reflect.Slice:
 			if canNumber(indexType.Kind()) {
 				return baseExpression.Index(int(castInt64(indexExpression)))
@@ -1086,7 +1091,7 @@ func (st *Runtime) evalBaseExpressionGroup(node Node) reflect.Value {
 		// limit the number of pointers to follow
 		for dereferenceLimit := 2; resolved.Kind() == reflect.Ptr && dereferenceLimit >= 0; dereferenceLimit-- {
 			if resolved.IsNil() {
-				return reflect.ValueOf("")
+				return emptyString
 			}
 			resolved = reflect.Indirect(resolved)
 		}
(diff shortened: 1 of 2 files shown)


eval.go
```diff
@@ -1445,15 +1450,18 @@ func castInt64(v reflect.Value) int64 {
 var cachedStructsMutex = sync.RWMutex{}
 var cachedStructsFieldIndex = map[reflect.Type]map[string][]int{}
 
-func getFieldOrMethodValue(key string, v reflect.Value) reflect.Value {
+func getFieldOrMethodValue(key string, v reflect.Value) (r reflect.Value) {
```
annismckenzie: You're not using `r` anywhere? 🤔

eval_test.go
```diff
@@ -185,6 +185,7 @@ func TestEvalActionNode(t *testing.T) {
 	RunJetTest(t, data, nil, "actionNode_MultAdd", `{{ 2+4*4 }}`, fmt.Sprint(2+4*4))
 	RunJetTest(t, data, nil, "actionNode_MultAdd1", `{{ 4*2+4 }}`, fmt.Sprint(4*2+4))
 	RunJetTest(t, data, nil, "actionNode_MultAdd2", `{{ 2+4*2+4 }}`, fmt.Sprint(2+4*2+4))
+	RunJetTest(t, data, nil, "actionNode_MultAdd3_6not8", `{{2+2*2}}`, fmt.Sprint(2+2*2))
```
annismckenzie: Whats's this test for?

annismckenzie:
I don't quite understand the side effect you mention in the description. Could you clarify? 😅

vetcher:
It's all about error message, that may be confusing for somebody.
Look at my example:
B is an interface; accessing .A.B.C should return _there is no field or method C in_ **interface{}**, but it will return _there is no field or method C in_ **string** because error message uses returned type of `getFieldOrMethodValue` function, not real `nil`.

vetcher:
Btw, may be it's time to add something like [that](https://github.com/golang/go/blob/master/src/text/template/exec.go#L617)?

eval.go
```diff
@@ -602,7 +603,11 @@ func (st *Runtime) evalPrimaryExpressionGroup(node Expression) reflect.Value {
 					node.errorf("%s is not assignable|convertible to map key %s", indexType.String(), key.String())
 				}
 			}
-			return baseExpression.MapIndex(indexExpression)
+			baseExpression = baseExpression.MapIndex(indexExpression)
+			if baseExpression.Kind() == reflect.Interface && baseExpression.IsNil() {
+				return emptyString
```
sauerbraten: I don't think it's a good idea to return a string here regardless of the underlying map's value type.

IMO, when a map contains an actual `nil` value for a key, then that value should be returned as-is to the caller.

We use map access syntax a lot to work on JSON objects (deserialized into map[string]interface{}), and checking `m["someKey"] == nil` is a lot clearer when expecting a `null` in the original JSON data than `m["someKey"] == ""` or `isset(m["someKey"]` (which would also be false if the key doesn't exist in the map at all).

When a map doesn't contain the indexExpression as key, `baseExpression.MapIndex(indexExpression)` will return `reflect.Value{}` and either that should be returned, to signal to the runtime that the key was not found, or the zero value of the map's value type should be returned (`reflect.Zero(baseExpression.Type().Elem())`), but then the runtime can never be sure if the value came from the map or if the key didn't exist.

sauerbraten: Returning the zero value of the map's value type means we can call methods etc. without any error when the map is a `map[string]User` (i.e. value type is a custom struct type), which could hide/shadow missing data.

annismckenzie: Got it, sounds very reasonable when you explain it like that. 😅

annismckenzie:
Closing because of the discussions in https://github.com/CloudyKit/jet/pull/126#issuecomment-573036224.

PR #108, opened 2019-01-18
YieldTemplate panics with exact error if GetTemplate errored

SVyatoslavG:
Added error output on missing template during GetTemplate call.


jhsx:
Hi @SVyatoslavG, thanks for the contribution.

tooolbox:
@jhsx Seems like a pretty trivial and beneficial change.  What do we think about merging before we hit the 1-year mark?

annismckenzie:
You're right, this looks good to me. Just as a note: because of #122 and #123 the master branch is broken right now. I can merge it but I won't release a new 2.x release just yet.

PR #112, opened 2019-03-20
make Ranger return interface{}, not reflect.Value

sauerbraten:
Previously, the key/index and element values returned by Ranger were reflect.Values pointing to the underlying data (see https://golang.org/src/reflect/value.go?s=67137:67170#L2242 and https://golang.org/src/reflect/value.go#L140). This meant you could not compare e.g. elements of a string slice to a string literal. Also, storing a copy of a certain slice element's index and using it outside of the range did not work, since the copy only pointed to the actual index (which had the value len+1 after the range).

### DIFF
--- a/default.go
+++ b/default.go
@@ -65,7 +65,7 @@ func init() {
 			a.RequireNumOfArguments("len", 1, 1)
 
 			expression := a.Get(0)
-			if expression.Kind() == reflect.Ptr {
+			if expression.Kind() == reflect.Ptr || expression.Kind() == reflect.Interface {
 				expression = expression.Elem()
 			}
 
--- a/eval.go
+++ b/eval.go
@@ -668,7 +668,7 @@ func (st *Runtime) isSet(node Node) bool {
 		indexExpression := st.evalPrimaryExpressionGroup(node.Index)
 
 		indexType := indexExpression.Type()
-		if baseExpression.Kind() == reflect.Ptr {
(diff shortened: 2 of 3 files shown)


sauerbraten:
https://github.com/CloudyKit/jet/pull/112/commits/21e5963e201a5b4a27e14b8d2ee06f55a3300566 fixes a bug in how isset() evaluated field access expressions: because of variable shadowing in a loop (`value := getFieldOrMethodValue(node.Field[i], value)`), all fields were accessed on the base instead of the preceding field. This meant accessing a first-level field worked as expected, but a second-level field only worked when a field with the same name also happened to exist on the base of the access expression.

Similar code without this bug existed in evalBaseExpressionGroup(). To prevent the two implementations from diverging, the code was moved into a new evalFieldAccessExpression() function. In isSet(), the error is ignored so as to not stop template execution when visiting a non-existent field.

The test "IsSetExpression_3" added to eval_test.go failed before this fix.

annismckenzie:
What a long time it's been. 😞 These changes make sense. Because they change the `Ranger` interface, I'll have to bump the version to v3 (and add go modules along the way).

sauerbraten:
I saw I pushed a few more changes to the branch later, which are unrelated to this PR. Most notably, the behavior of `isset` changes to return false when a variable is `nil`, even if set explicitly to that value.

annismckenzie:
I'll merge this, thanks! I'll fix the docs and examples using the Ranger interface in a separate pull request.

PR #121, opened 2019-12-30
Add Go modules and use v3

annismckenzie:
Because of upcoming backwards-incompatible changes I went ahead and used `v3` while adding Go modules.


go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/CloudyKit/jet/v3
+
+go 1.12
+
+require github.com/CloudyKit/fastprinter v0.0.0-20170127035650-74b38d55f37a
+
+replace github.com/CloudyKit/fastprinter => github.com/annismckenzie/fastprinter v0.0.0-20180721091441-b9f2fb63ce50 // PR 1 at https://github.com/CloudyKit/fastprinter/pull/1
```
annismckenzie: That pull request still isn't merged so I had to use replace…

annismckenzie:
This is done, no code changes other than introducing modules and using v3 – merging.

PR #124, opened 2020-01-08
Revert #96

### LINKED ISSUE #123: Regression: Unexpected dereference of returned struct pointer
I've observed a regression caused by https://github.com/CloudyKit/jet/commit/91ac9cb0201cc9d6b4130277c2f7e4b3685850f6 (#97)

### Summary
For this template:
```handlebars
{{apple := GetAppleByName("honeycrisp")}}
{{TellFlavor(apple)}}
```

With this code:
```go
func main() {

	// ...

	apples := map[string]*Apple{
		"honeycrisp": {
			Flavor: "crisp",
		},
		"red-delicious": {
			Flavor: "poor",
		},
		"granny-smith": {
			Flavor: "tart",
		},
	}

	// ...

	vars.SetFunc("GetAppleByName", func(a jet.Arguments) reflect.Value {
		name := a.Get(0).String()
		return reflect.ValueOf(apples[name])
	})

	vars.SetFunc("TellFlavor", func(a jet.Arguments) reflect.Value {
		apple := a.Get(0).Interface().(*Apple)
		flav := apple.GetFlavor()
		return reflect.ValueOf(flav)
	})

	// ...
}

type Apple struct {
	Flavor string
}

func (a *Apple) GetFlavor() string {
	return a.Flavor
}
```

The expected output would be `crisp` and this worked fine with v2.1.2

After https://github.com/CloudyKit/jet/commit/91ac9cb0201cc9d6b4130277c2f7e4b3685850f6 the output is:
```
panic: interface conversion: interface {} is main.Apple, not *main.Apple
// stack trace elided
```

### Example Project
https://github.com/tooolbox/jet-example has two branches that cover this:
- v2.1.2-pointers (working)
- v2.1.3-91ac9cb (broken)

### Solution
It seems as though `{{apple := GetAppleByName("honeycrisp")}}` is returning a `*main.Apple` but the new code is dereferencing that pointer which is unexpected.  I can see where this might be handy for a `*string` or `*int` but for a `*struct{}` I can't think why this would be the desired behavior.

tooolbox:
Fixes #123 

This is a draft because we need a test to cover this scenario and I don't feel extremely qualified to write it, so I'm hoping someone else can 😃   But I at least wanted to start the ball rolling.

(Evidence that this handles #123 is visible [on this branch](https://github.com/tooolbox/jet-example/tree/v2.1.3-pointers-fixed).)

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -16,13 +16,12 @@ package jet
 
 import (
 	"fmt"
+	"github.com/CloudyKit/fastprinter"
 	"io"
 	"reflect"
 	"runtime"
 	"strconv"
 	"sync"
-
-	"github.com/CloudyKit/fastprinter"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


tooolbox:
This has been covered elsewhere.

PR #125, opened 2020-01-09
Fix dereferencing printable types

sarge:
Fixes I hope the case where a non printable type - struct is being returned.

Apologies for the build on the modules branch, but that is how I do all development now.


annismckenzie:
I merged #121 just now – can you rebase this, please? :)

eval.go
```diff
@@ -1091,14 +1091,32 @@ func (st *Runtime) evalBaseExpressionGroup(node Node) reflect.Value {
...
+			case reflect.String,
+				reflect.Bool,
+				reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
+				reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
+				reflect.Float32, reflect.Float64:
+				return rTest
+			}
+
+			if dereferenceLimit == 0 {
+				return reflect.ValueOf("<nil>")
```
sauerbraten: doesn't this make it impossible to ever do anything at all with, say, a `***MyStruct`?

I mean, hopefully never needs to have three levels of indirection, but we shouldn't make it impossible to work with those. I'd say we return the original value in this case. This function is not only used to render values into the output string, but also to evaluate the base expression in chain expressions, which is usually an identifier, meaning this code will run when evaluating `{{a.Name()}}` where `a` might be a pointer to a variable of `type A **Foo` or something like that.

sauerbraten:
I thought about it a bit more, and I actually believe that template evaluation is the wrong place for implicit pointer dereferentiation. Instead, this should be done when printing the result of an evaluation, i.e. in this case in https://github.com/CloudyKit/fastprinter. I will prepare a pull request there.

tooolbox:
> I thought about it a bit more, and I actually believe that template evaluation is the wrong place for implicit pointer dereferentiation. Instead, this should be done when printing the result of an evaluation, i.e. in this case in https://github.com/CloudyKit/fastprinter.

You know, that had occurred to me actually.  So I guess this means that if we merge https://github.com/CloudyKit/fastprinter/pull/2 then this PR should remove the indirection entirely, no?

sarge:
Great idea, makes much better sense to me too. I was a bit iffy about my approach to begin with.

PR #126, opened 2020-01-09
Misc fixes related to Ranger and pointer evaluation/printing

### LINKED ISSUE #117: Context leak on recover
I noticed a fix given in [this fork](https://github.com/kabukky/jet/commit/6f7ec1b8f0c603d6740d016409a230c57829b100) for a [line in eval.go](https://github.com/CloudyKit/jet/blob/master/eval.go#L224) indicating that this potentially leaks variables between executions.

This should be reviewed for severity and potentially merging here.

### LINKED ISSUE #122: Regression: Ranging over slice from Fast Function
I've observed a regression caused by https://github.com/CloudyKit/jet/commit/6d666f94dfe73004a23e8e41d7532ac8602c9132 (#112)

### Summary
For this template:
```handlebars
{{sorted := SortApples(myApples)}}
{{range i, apple := sorted}}
	{{i}}: {{apple.GetFlavor()}}
{{end}}
```

(Where `SortApples` is a [fast function](https://github.com/CloudyKit/jet/wiki/5.-Advanced-Customization#fast-functions) aka [jet.Func](https://godoc.org/github.com/CloudyKit/jet#Func).)

Jet was able to range over the returned slice up until https://github.com/CloudyKit/jet/commit/6d666f94dfe73004a23e8e41d7532ac8602c9132 at which point the template execution would error with: `there is no field or method "GetFlavor" in main.Apple`

### Example Project
https://github.com/tooolbox/jet-example has two branches:
- v2.1.2 (working)
- v2.1.3-6d666f94dfe7 (broken)

-------------------------------------

I understand we're now going for a v3 with these commits, so breaking changes might be expected, but this seems like it would hurt fast functions a lot.

cc @sauerbraten

### LINKED ISSUE #123: Regression: Unexpected dereference of returned struct pointer

sauerbraten:
Sorry for the big PR, but this was the easiest way to clean up the mess I created with the last PR :D

- make `Ranger` return `reflect.Values` again (see https://github.com/CloudyKit/jet/issues/122#issuecomment-572311118)
- fix #117 (using fix from https://github.com/CloudyKit/jet/issues/117#issuecomment-518576462)
- fix #122 (@tooolbox: check out https://github.com/sauerbraten/jet-example/tree/v2.1.3-6d666f94dfe7)
- fix #123 (@tooolbox: check out https://github.com/sauerbraten/jet-example/tree/v2.1.3-91ac9cb)
- move pointer dereferencing to fastprinter package (see https://github.com/CloudyKit/jet/pull/125#issuecomment-572555115, https://github.com/CloudyKit/fastprinter/pull/2)
- add `IsSet` function to `*Arguments` type

This PR obsoletes #125.

### DIFF
--- a/eval_test.go
+++ b/eval_test.go
@@ -373,11 +373,19 @@ func TestEvalSliceExpression(t *testing.T) {
 	RunJetTest(t, nil, []string{"111"}, "SliceExpressionSlice_IfLen", `{{if len(.) > 0}}{{.[0]}}{{end}}`, `111`)
 }
 
+type StringerType struct {
+	//
+}
+
+func (st *StringerType) String() string {
+	return "StringerType implements fmt.Stringer"
+}
+
 func TestEvalPointerExpressions(t *testing.T) {
 	var data = make(VarMap)
 	var s *string
 	data.Set("stringPointer", s)
-	RunJetTest(t, data, nil, "StringPointer_1", `{{ stringPointer }}`, "")
+	RunJetTest(t, data, nil, "StringPointer_1", `{{ stringPointer }}`, "<nil>")
 
 	s2 := "test"
 	data.Set("stringPointer2", &s2)
@@ -389,7 +397,30 @@ func TestEvalPointerExpressions(t *testing.T) {
 
 	var i *int
 	data.Set("intPointer", &i)
-	RunJetTest(t, data, nil, "IntPointer_i", `{{ intPointer }}`, "")
+	RunJetTest(t, data, nil, "IntPointer_i", `{{ intPointer }}`, "<nil>")
+
+	st := StringerType{}
+	data.Set("stringerType", &st) // stringerType won't be dereferenced
+	RunJetTest(t, data, nil, "StringerTypePointer", `{{ stringerType }}`, "StringerType implements fmt.Stringer")
+
+	st2 := &StringerType{}
+	data.Set("stringerType2", &st2) // stringerType2 will be dereferenced just once
+	RunJetTest(t, data, nil, "StringerTypePointer2", `{{ stringerType2 }}`, "StringerType implements fmt.Stringer")
+
+	u := User{
+		Name:  "Pablo Escobbar",
+		Email: "pablo.escobar@cartel.mx",
+	}
+	data.Set("user", &u) // user will be dereferenced once
+	RunJetTest(t, data, nil, "UserPointer", `{{ user }}`, "{Pablo Escobbar pablo.escobar@cartel.mx}")
+
+	u1 := &u
+	data.Set("user1", &u1) // user1 will be dereferenced twice
+	RunJetTest(t, data, nil, "UserPointer1", `{{ user1 }}`, "{Pablo Escobbar pablo.escobar@cartel.mx}")
+
+	u2 := &u1
+	data.Set("user2", &u2) // user2 will be dereferenced only twice
+	RunJetTest(t, data, nil, "UserPointer2", `{{ user2 }}`, "&{Pablo Escobbar pablo.escobar@cartel.mx}")
 }
 
 func TestEvalPointerLimitNumberOfDereferences(t *testing.T) {
@@ -397,15 +428,56 @@ func TestEvalPointerLimitNumberOfDereferences(t *testing.T) {
 
 	var i *int
 	data.Set("intPointer", &i)
-	RunJetTest(t, data, nil, "IntPointer_i", `{{ intPointer }}`, "")
+	RunJetTest(t, data, nil, "IntPointer_i", `{{ intPointer }}`, "<nil>")
 
 	j := &i
 	data.Set("intPointer", &j)
-	RunJetTest(t, data, nil, "IntPointer_j", `{{ intPointer }}`, "")
+	RunJetTest(t, data, nil, "IntPointer_j", `{{ intPointer }}`, "<nil>")
 
-	k := &j
-	data.Set("intPointer", &k)
-	RunJetTest(t, data, nil, "IntPointer_1", `{{ intPointer }}`, "<nil>")
+	// k := &j
+	// data.Set("intPointer", &k)
+	// RunJetTest(t, data, nil, "IntPointer_1", `{{ intPointer }}`, "<nil>")
+}
+
+type Apple struct {
+	Flavor string
+}
+
+func (a *Apple) GetFlavor() string {
+	return a.Flavor
+}
+
+func (a *Apple) GetFlavorPtr() *string {
+	return &a.Flavor
+}
+
+func TestApple(t *testing.T) {
+
+	apples := map[string]*Apple{
+		"honeycrisp": {
+			Flavor: "crisp",
+		},
+		"red-delicious": {
+			Flavor: "poor",
+		},
+		"granny-smith": {
+			Flavor: "tart",
+		},
+	}
+
+	var data = make(VarMap)
+	data.SetFunc("GetAppleByName", func(a Arguments) reflect.Value {
+		name := a.Get(0).String()
+		return reflect.ValueOf(apples[name])
+	})
+
+	data.SetFunc("TellFlavor", func(a Arguments) reflect.Value {
+		apple := a.Get(0).Interface().(*Apple)
+		flav := apple.GetFlavor()
+		return reflect.ValueOf(flav)
+	})
+
+	RunJetTest(t, data, nil, "LookUpApple", `{{apple := GetAppleByName("honeycrisp")}}{{TellFlavor(apple)}}`, "crisp")
 }
 
 func TestEvalStructFieldPointerExpressions(t *testing.T) {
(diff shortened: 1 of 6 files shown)


sauerbraten:
Oh, right: you probably want to wait before merging this until https://github.com/CloudyKit/fastprinter/pull/2 is merged and the replace directive for fastprinter is unnecessary: https://github.com/CloudyKit/jet/pull/126/files#diff-37aff102a57d3d7b797f152915a6dc16

tooolbox:
This looks fantastic :)

go.mod
```diff
@@ -4,4 +4,4 @@ go 1.12
 
 require github.com/CloudyKit/fastprinter v0.0.0-20170127035650-74b38d55f37a
 
-replace github.com/CloudyKit/fastprinter => github.com/annismckenzie/fastprinter v0.0.0-20180721091441-b9f2fb63ce50 // PR 1 at https://github.com/CloudyKit/fastprinter/pull/1
+replace github.com/CloudyKit/fastprinter => github.com/sauerbraten/fastprinter v0.0.0-20200109142611-26ce1e0b7924
```
annismckenzie: We can drop this now that those PRs have been merged. Can you also update the library above to the new commit?

node.go
```diff
@@ -61,7 +61,7 @@ func (node *NodeBase) error(err error) {
 }
 
 func (node *NodeBase) errorf(format string, v ...interface{}) {
-	panic(fmt.Errorf("Jet Runtime Error(%q:%d): %s", node.TemplateName, node.Line, fmt.Sprintf(format, v...)))
+	panic(fmt.Errorf("jet runtime error @ %q:%d: %s", node.TemplateName, node.Line, fmt.Sprintf(format, v...)))
```
annismckenzie: I don't think there's a reason to change this? The previous version was fine. :)

annismckenzie:
Wow, so many issues that'll be fixed in a single PR! ❤️ Technically, we didn't even have to bump the version to 3. 🤔 On the other hand, before I added Go modules, we weren't compatible with a v2 module anyway because the import paths needed to be changed. Anyways, I think we're good.

go.mod
sauerbraten: Done in https://github.com/CloudyKit/jet/pull/126/commits/ab0cc7f8326203963f1786c74a0c21332fab2411 👍

node.go
sauerbraten: True, I changed it to lowercase to be similar to usual Go error messages, but it's not needed. https://github.com/CloudyKit/jet/pull/126/commits/a04ce10e0ad8891da46a2d625248df24e58133c4

eval_test.go
```diff
@@ -389,23 +397,87 @@ func TestEvalPointerExpressions(t *testing.T) {
...
 	data.Set("intPointer", &i)
-	RunJetTest(t, data, nil, "IntPointer_i", `{{ intPointer }}`, "")
+	RunJetTest(t, data, nil, "IntPointer_i", `{{ intPointer }}`, "<nil>")
 
 	j := &i
 	data.Set("intPointer", &j)
-	RunJetTest(t, data, nil, "IntPointer_j", `{{ intPointer }}`, "")
+	RunJetTest(t, data, nil, "IntPointer_j", `{{ intPointer }}`, "<nil>")
+
+	// k := &j
```
annismckenzie: Can you drop these? This case is already tested in the previous test.

annismckenzie:
I'm still not a 100% on printing `<nil>` for nil pointers – see #103, #100 and #97. Are we okay with this?

eval_test.go
sauerbraten: This last test are supposed to show that a `***int` won't be resolved all the way to `<nil>` and instead would print as a pointer address, but there's no way to predict what that address will be. I can take them out of course.

sauerbraten:
> I'm still not a 100% on printing `<nil>` for nil pointers – see #103, #100 and #97. Are we okay with this?

fastprinter's PrintValue method has always printed `<nil>` for nil pointers or interfaces due to this line: https://github.com/CloudyKit/fastprinter/blob/74b38d55f37af5d6c05ca11147d616b613a3420e/printers.go#L202
https://github.com/CloudyKit/fastprinter/pull/2 didn't change that behavior.

In #100 & #97, pointer dereferencing was added to an `eval` function, which @sarge, @tooolbox and I agreed was a mistake in #125.
I wasn't aware of #103 but commented on it here: https://github.com/CloudyKit/jet/pull/103/files#r365232011. It boils down to the same reasoning: don't switch to string type in the middle of evaluation.

tooolbox:
> I'm still not a 100% on printing <nil> for nil pointers – see #103, #100 and #97. Are we okay with this?

I think @sauerbraten has the right of it; we don't want to implicitly switch to a string type.

For maps you can achieve an empty-string-if-nil by [doing this](https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax#maps):
```handlebars
{{ if b, ok := a["b"]; ok }}
  {{ b }}
{{ end }}
```
(I never actually use dot-syntax to access maps, so I'm not sure if you can do the two-value return in that scenario, but I imagine so.)

You could also [pipeline your output](https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax#pipelining) with a custom function to turn nil into an empty string, and in that case it's explicit.

Or if we felt like getting creative, could steal some operators from Swift:
```handlebars
<!-- nil-coalescing -->
{{ .A ?? "" }}

<!-- optional-unwrapping -->
{{ .A? }} <!-- outputs "" if nil -->
{{ .A?.B? }} <!-- chain it if desired -->
```
Anyway, that's a whole ball of wax but my concept is that it would be explicit, whatever it is.

annismckenzie:
> > I'm still not a 100% on printing  for nil pointers – see #103, #100 and #97. Are we okay with this?
> 
> I think @sauerbraten has the right of it; we don't want to implicitly switch to a string type.
> 
> For maps you can achieve an empty-string-if-nil by [doing this](https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax#maps):
> 
> ```
> {{ if b, ok := a["b"]; ok }}
>   {{ b }}
> {{ end }}
> ```
> 
> (I never actually use dot-syntax to access maps, so I'm not sure if you can do the two-value return in that scenario, but I imagine so.)
> 
> You could also [pipeline your output](https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax#pipelining) with a custom function to turn nil into an empty string, and in that case it's explicit.
> 
> Or if we felt like getting creative, could steal some operators from Swift:
> 
> ```
> 
> {{ .A ?? "" }}
> 
> 
> {{ .A? }} 
> {{ .A?.B? }} 
> ```
> 
> Anyway, that's a whole ball of wax but my concept is that it would be explicit, whatever it is.

All good, very astute observations!

eval_test.go
annismckenzie: Yeah, I'd drop the commented lines. After that, we can merge this!

sauerbraten: Ah, completely forgot about this: 
f30e9c7

annismckenzie:
Thank you for all the hard work – merging! 🎉

Issue #128, opened 2020-01-15
OSFileSystemLoader shouldn't manage multiple directories


sauerbraten:
Having `OSFileSystemLoader` manage multiple directories can lead to unexpected behavior:
- assume we have
    - `/home/bla/templates/foo/a/b/c.jet`
    - `/home/bla/templates/bar/a/b/c.jet`
- using `NewHTMLSet("/home/bla/templates/foo", "/home/bla/templates/bar")`
- the following is undefined: `{{ include "a/b/c.jet" }}`

Which template is included depends on the order in which you pass the directory paths to `NewHTMLSet`, which is documented nowhere.

tooolbox:
> the following is undefined: `{{ include "a/b/c.jet" }}`

Well, it's not so much undefined but rather it depends on the sequence of your directory paths 😆 

But your point is taken and I think this will make the API more obvious moving forward.

sauerbraten:
Okay, let's say *non-obivous ;)

sauerbraten:
Fixed by https://github.com/CloudyKit/jet/pull/129.

PR #129, opened 2020-01-15
Clean up template loading

### LINKED ISSUE #127: include & extend don't handle absolute paths correctly
Assuming I have a template at `foo/baz.jet`, `{{ include "foo/bar.jet" }}` will make `resolveNameSibling` try to resolve `foo/foo/bar.jet` before resolving `foo/bar.jet`: https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L153

`name` will be `"foo/bar.jet"`, `sibling` will be `"foo/baz.jet"`.

I would like to simplify template resolving & loading and could fix this issue at the same time. Seeing as we are making breaking changes with v3 anyway, I would like to make jet handle paths like most software does:
- `./asd.jet`, `asd.jet`, `../asd.jet` will be treated as relative paths and will be resolved relative to the current template's path
- `/foo.jet`, `/foo/bar.jet` will be treated as absolute paths and will be looked up directly in the templates cache and/or the loader, without using the current template's path

Do you want me to proceed with this? Did I miss something?

sauerbraten:
This fixes #127 (paths in templates will be handled as described there) and #128.

Breaking changes:

- `OSFileSystemLoader` only supports a single directory instead of multiple
- the `AddPath` and `AddGopathPath` functions were removed
- ~~`Loader.Exists` no longer returns the "real" file name~~
- after `include` or `extends`, relative paths will no longer be looked up as absolute paths (see #127)

I removed the ability to have multiple directories managed by a single `OSFileSystemLoader` becaue of #128, and because I feel like
1. most users will have all templates in a single directory, organized into subdirectories there,
2. with the new handling of relative paths, the top directory name will appear in paths less often, and
3. you can always use `loaders.Multi` to get back the functionality a single `OSFileSystemLoader` provided before

I added a note to `loaders.Multi` warning that lookup order depends on the order in which loaders are passed in during instantiation.

### DIFF
--- a/constructors.go
+++ b/constructors.go
@@ -21,133 +21,133 @@ import (
 )
 
 func (t *Template) newSliceExpr(pos Pos, line int, base, index, len Expression) *SliceExprNode {
-	return &SliceExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeSliceExpr, Pos: pos, Line: line}, Index: index, Base: base, EndIndex: len}
+	return &SliceExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeSliceExpr, Pos: pos, Line: line}, Index: index, Base: base, EndIndex: len}
 }
 
 func (t *Template) newIndexExpr(pos Pos, line int, base, index Expression) *IndexExprNode {
-	return &IndexExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeIndexExpr, Pos: pos, Line: line}, Index: index, Base: base}
+	return &IndexExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeIndexExpr, Pos: pos, Line: line}, Index: index, Base: base}
 }
 
 func (t *Template) newTernaryExpr(pos Pos, line int, boolean, left, right Expression) *TernaryExprNode {
-	return &TernaryExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeTernaryExpr, Pos: pos, Line: line}, Boolean: boolean, Left: left, Right: right}
+	return &TernaryExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeTernaryExpr, Pos: pos, Line: line}, Boolean: boolean, Left: left, Right: right}
 }
 
 func (t *Template) newSet(pos Pos, line int, isLet, isIndexExprGetLookup bool, left, right []Expression) *SetNode {
-	return &SetNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeSet, Pos: pos, Line: line}, Let: isLet, IndexExprGetLookup: isIndexExprGetLookup, Left: left, Right: right}
+	return &SetNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeSet, Pos: pos, Line: line}, Let: isLet, IndexExprGetLookup: isIndexExprGetLookup, Left: left, Right: right}
 }
 
 func (t *Template) newCallExpr(pos Pos, line int, expr Expression) *CallExprNode {
-	return &CallExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeCallExpr, Pos: pos, Line: line}, BaseExpr: expr}
+	return &CallExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeCallExpr, Pos: pos, Line: line}, BaseExpr: expr}
 }
 func (t *Template) newNotExpr(pos Pos, line int, expr Expression) *NotExprNode {
-	return &NotExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeNotExpr, Pos: pos, Line: line}, Expr: expr}
+	return &NotExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeNotExpr, Pos: pos, Line: line}, Expr: expr}
 }
 func (t *Template) newNumericComparativeExpr(pos Pos, line int, left, right Expression, item item) *NumericComparativeExprNode {
-	return &NumericComparativeExprNode{binaryExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeNumericComparativeExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
+	return &NumericComparativeExprNode{binaryExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeNumericComparativeExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
 }
 
 func (t *Template) newComparativeExpr(pos Pos, line int, left, right Expression, item item) *ComparativeExprNode {
-	return &ComparativeExprNode{binaryExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeComparativeExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
+	return &ComparativeExprNode{binaryExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeComparativeExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
 }
 
 func (t *Template) newLogicalExpr(pos Pos, line int, left, right Expression, item item) *LogicalExprNode {
-	return &LogicalExprNode{binaryExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeLogicalExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
+	return &LogicalExprNode{binaryExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeLogicalExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
 }
 
 func (t *Template) newMultiplicativeExpr(pos Pos, line int, left, right Expression, item item) *MultiplicativeExprNode {
-	return &MultiplicativeExprNode{binaryExprNode{NodeBase: NodeBase{TemplateName: t.Name, NodeType: NodeMultiplicativeExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
+	return &MultiplicativeExprNode{binaryExprNode{NodeBase: NodeBase{TemplatePath: t.Name, NodeType: NodeMultiplicativeExpr, Pos: pos, Line: line}, Operator: item, Left: left, Right: right}}
 }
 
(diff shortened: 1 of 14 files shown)


tooolbox:
> most users will have all templates in a single directory, organized into subdirectories there

For the record, I have many projects which have templates in multiple directories.  😄 

> you can always use loaders.Multi to get back the functionality a single OSFileSystemLoader provided before

Do we have a test to ensure these two things are precisely equivalent?  (I didn't see one, but I might have missed it.)

> I added a note to loaders.Multi warning that lookup order depends on the order in which loaders are passed in during instantiation.

Nice!  Probably could use the same note on the existing OSFileSystemLoader and cousins if this isn't immediately merged.

> `Loader.Exists` no longer returns the "real" file name

Could you explain the motivation for this?  I have code that uses this.

> Breaking changes:

I feel like we need to have some versions tagged before there's further breaking changes.  See for example https://github.com/getsentry/sentry-go/issues/135

Specifically, I'd like to get a v3 version with Go Modules support cut & tagged.  Then maybe this sort of thing can go in v4.  What do you think?

sauerbraten:
> > you can always use loaders.Multi to get back the functionality a single OSFileSystemLoader provided before

> Do we have a test to ensure these two things are precisely equivalent? (I didn't see one, but I might have missed it.)

I can't write a test comparing functionality from two different revisions, sorry 🙃
If you look at the code of the current `OSFileSystemLoader` you'll see it has the same loop that has always been in `loaders.Multi`:
- current `OSFileSystemLoader`: https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/loader.go
- current `loaders.Multi`: https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/loaders/multi/multi.go
- new `loaders.Multi`: https://github.com/sauerbraten/jet/blob/clean-up-tmpl-loading/loaders/multi/multi.go

> > Loader.Exists no longer returns the "real" file name

> Could you explain the motivation for this? I have code that uses this.

I removed it because inside jet, there's no usage of `Exists` that isn't (or shouldn't be) immediately followed by an `Open` (as long as `Exists` returned `true`):
- https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L196 is followed by https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L197
- https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L203 is followed by https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L204
- https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L228 is followed by https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L230
- https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L242 is followed by a lookup from cache (which the new `getTemplate` tries first, too) or by https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L263 which is followed by a lookup from cache (again, same behavior with the new `getTemplate`) or https://github.com/CloudyKit/jet/blob/33cfc27b3e00072655fdb3af24c3253ea5bffb8f/template.go#L273

So I built the new `GetTemplate` and `getTemplate` functions to return the requested template if found. They use `Exists` and `Open` right now to try the possible file extensions, but I could imagine getting rid of `Exists` entirely to save a syscall. When `Open` fails, move to the next extension, when all failed, it doesn't exist.

Maybe there's a good reason to provide an `Exists` method through jet's `Loader` interface. Maybe even a reason to return the "corrected" template path. What's your use case?

> I feel like we need to have some versions tagged before there's further breaking changes.

I agree! It doesn't matter which major version includes fixes for #127 and #128, as long as they are fixed :)

sauerbraten:
From my last comment:

> I could imagine getting rid of `Exists` entirely to save a syscall. When Open fails, move to the next extension, when all failed, it doesn't exist.

Turns out that's a bad idea: opening a file could return many kinds of errors (for example, permissions). When a template exists but can't be read, we should exit the loop and return the error from `Open`.

`Exists` should stay. Still, why should it guess the correct file from an incomplete path? We could make this its own function if needed.

tooolbox:
> I can't write a test comparing functionality from two different revisions, sorry

Doh.  (What, nobody taught you how to do that?  C'mon... 😛 )  Well, point taken.  But maybe tests should exist for the current version and then we can see that the new revision doesn't break them.  I know that's wishful and hand-wavy but that's the best I got right now.

> I removed it because inside jet, there's no uses of resolveNameSibling (the only user of Exists, via resolveName) that aren't followed (as long as Exists returned true) by a cache lookup or a call to loadFromFile (the only user of Open)

Sure, inside Jet.  But given that it's public API there may be code that depends on this.

> Maybe there's a good reason to provide an Exists method through jet's Loader interface. Maybe even a reason to return the "corrected" template path. What's your use case?

Using the returned path to perform other operations on the files.  Using the jet loader as a source of truth about what files exist within its zone of responsibility.  Using the paths returned by it for our own caching, logging, or indexing purposes.  Plus creating interface of `{ Exists, Open }` as a generic loading mechanism which is fulfilled by several things, only one of which is a Jet loader.

> Exists should stay. Still, why should it guess the correct file from an incomplete path? We could make this its own function if needed.

I figure that "guessing the correct file from an incomplete path" is part of why `Exists()` returns the path, so you know what it wound up with.  Maybe if you don't want the template name "resolution", you could make a `StrictLoader`?

Hm, actually, which part are you referring to as guessing the correct file?  Looking at the existing `OSFileSystemLoader` it's just appending what you give it to its path(s), which doesn't seem like much guesswork.  (Sorry if I'm asking dumb questions, I'm obviously not as deep in the codebase as you are.)

----------------------------

Suggest pulling the include/extend handling as per #127 into their own PR.

sauerbraten:
> there may be code that depends on this.

Which is why I included in under "Breaking changes" :) I know this means users will have to adapt, but I believe it's worth it. It was mainly useful when the FS loader was managing multiple directories. With the new loader, you can only have a single directory, so when the loader for that directory says the path exists within its directory, `filepath.Join(loaderDir, path)` will give you what you got from the old `Exists` method.

>  Using the jet loader as a source of truth about what files exist within its zone of responsibility.

The "new" loader will still tell you wether a path exists in its directory or not.

> Using the paths returned by it for our own caching, logging, or indexing purposes.

The complete path is easy to obtain, see above. Maybe the multi loader could have a `Resolve` method that tries all the loaders inside it, so you don't have to do this manually, but IMO it should not be part of the loader interface.

> interface of { Exists, Open } as a generic loading mechanism

To be honest, I'll say that that's not the job of the jet codebase. Users can easily create their own interface with a resolving `Exists` method, where they wrap the jet loaders method with path resolution.

sauerbraten:
If these changes are too broad to merge them at once, I can split the PR up into multiple, but they would probably only work well in a certain order, since they all mostly touch the same file...

tooolbox:
> It was mainly useful when the FS loader was managing multiple directories. With the new loader, you can only have a single directory, so when the loader for that directory says the path exists within its directory, filepath.Join(loaderDir, path) will give you what you got from the old Exists method.

I could just as easily say that it's a convenient feature for the Loader to return the combined `filepath.Join(loaderDir, path)` rather than you having to do it yourself.  I don't see any benefit to removing that feature.

> The complete path is easy to obtain, see above. 

The `dir` var of `OSFileSystemLoader` isn't currently exported.  You would have to keep a copy of that data somewhere to then call `filepath.Join` on, or if we exported it then you would have to know that you are dealing with an `OSFileSystemLoader` rather than the generic `Loader` interface, and then why do you have an interface at all.

> Maybe the multi loader could have a Resolve method that tries all the loaders inside it, so you don't have to do this manually, but IMO it should not be part of the loader interface.

See above; you then have to know it's a MultiLoader instead of just a Loader.

> I know this means users will have to adapt, but I believe it's worth it.

I think there are aspects of your PR which are worth it and aspects which are not.  The "undefined behavior" nature of `OSFileSystemLoader` is something that should be improved and I applaud that you're doing that.  Changing it to accept one path, and advising devs to use `MultiLoader` where they need several directories, is IMO a good idea in the long run.

However, changing the `Loader` interface is orthogonal to the issue you set out to address and doesn't bring any tangible benefits.

sauerbraten:
> The `dir` var of `OSFileSystemLoader` isn't currently exported. You would have to keep a copy of that data somewhere

It's a directory path, most if not all users of jet should have that as part of their configuration (since it's something that you should be able to pass in as env var, command line flag, or via a config file).

> See above; you then have to know it's a MultiLoader instead of just a Loader.

You instantiate it, so you *do* know that it's a multi loader.

> However, changing the Loader interface is orthogonal to the issue you set out to address and doesn't bring any tangible benefits.

This PR is about more than fixing #128, but I guess I can first fix the issues in separate PRs, and then we can look at this again.

I'd also appreciate feedback from the repo owners about this, of course!

tooolbox:
> It's a directory path, most if not all users of jet should have that as part of their configuration (since it's something that you should be able to pass in as env var, command line flag, or via a config file).

> You instantiate it, so you do know that it's a multi loader.

These are assumptions about the structure of other people's programs.  In my case, the wire-up of Jet is done far away from components that will need to use the path returned from `Exists`.  These components also have no knowledge of Jet itself (it is treated as one source of templates/pages out of several) and therefore don't know anything is a multi-loader.

> This PR is about more than fixing #128, but I guess I can first fix the issues in separate PRs, and then we can look at this again.

Looking above, your PR says it's for #128 and #127, and I see value in fixing both of these things.  Absolute path handling should be standard and clear.  `OSFileSystemLoader` can take a single path because there's a clear alternative in `MultiLoader`.

However, I would really appreciate it if we don't modify the `Loader` interface.  Like I said I don't see actual tangible benefits.  I can be wrong, that's fine, but currently it comes across as an aesthetics and code-style point, whereas I have production code that depends on this API.

sauerbraten:
@tooolbox I restored the old Exists() signature and functionality. I'd rather not split this PR into two, I feel it's okay to address both #127 and #128 at once, since they are kinda related (dealing with directories etc.) and also because I dread the extra work... :D 
This passes all tests, everything builds, it's a great candidate to merge before releasing v4 in my opinion. Did I miss any dealbreakers?

tooolbox:
Hey @sauerbraten and thanks for pinging me!

I just went through it and it seems good to me.  I see the `OSFileLoader` bit, sounds good, a little bit of renaming, some good cleanup.  The filepath handling stuff was a little deep for me to fully track but I remember our discussion and how it was a little odd so I trust your fixed implementation.

Great on the Exists() stuff.  :)  Yeah at this point obviously there are breaking changes but all good ones I think, so in my case I will be able to upgrade (not right away necessarily) and I think that will be true for most users.

So, LGTM!

PR #131, opened 2020-01-22
Add return keyword & exec() built-in function

sauerbraten:
This will allow returning a value from a template and using that value in the calling template, enabling users to have "function" templates they can reuse.

foo.jet:
```
{{ f := "f" }}
{{ o := "o" }}

... some content that will be discarded when this template runs inside exec() ...

{{ return f+o+o }}
```

test.jet:
```
{{ foo := exec("./foo") }}

<p>below the line, there will only be "foo":</p>
<hr>
{{ foo }}
```


annismckenzie:
Uuuuh, this looks neat! Can you add some unit tests?

sauerbraten:
Here you go!

testData/execReturn/foo.jet
```diff
@@ -0,0 +1,4 @@
+{{ f := "f" }}
+{{ o := "o" }}
+... some content that will be discarded when this template runs inside exec() ...
+{{ return f+o+o }}
```
annismckenzie: Nit: can you add a trailing newline?

testData/execReturn/test.jet
```diff
@@ -0,0 +1 @@
+{{ foo := exec("./foo") }}{{ foo }}
```
annismckenzie: In this file as well.

annismckenzie:
After we merge this it'd be nice if you added a couple lines of documentation to the wiki for this feature. I'll arrange access for you.

testData/execReturn/foo.jet
sauerbraten: 0085ed7 :+1:

testData/execReturn/test.jet
sauerbraten: 0085ed7 :+1:

testData/execReturn/foo.jet
annismckenzie: And now your tests are failing because of the added newlines. I guess we don't want those, right?

sauerbraten: We don't want the newlines or we don't want the test failures? :D
Anyway, I somehow had to change only one of the tests... is this expected behavior?
db4d56e

sauerbraten:
I'd love to, but I don't have permission to do that :D
No write access 🤷‍♂

annismckenzie:
I think it's time – there you go. 😬

sauerbraten:
Is there a policy whether to merge as-is or can/should I squash-merge?

annismckenzie:
Always use `Rebase and merge`, please. 👍

annismckenzie:
No squashing. I like keeping the history.

PR #138, opened 2020-02-27
Use key/fieldname/index resolution code from Go's text/template

sauerbraten:
With this PR, jet uses text/template's evalField() (with some changes) and a bit of text/template's index() to resolve fields (in maps and structs, by name) and indices (in slices, strings, arrays).

The goal is to rely on tried and tested logic from the standard library as much as possible. The new resolveIndex() function is also used for both, isset() checks and actual evaluation, reducing code size and complexity.

This PR slightly changes the truthiness table for value's in jet:

- empty, but non-nil, maps and slices are now considered "truthy" values (matching what you'd expect from an `isset(m)` call on such a value IMO)
- channels, functions and interfaces are now considered "truthy" when they are non-nil, while they were always "falsy" before
- numeric values are now only "falsy" when they are 0, while all values <= 0 were considered "falsy" before

This PR removes support for indexing into structs with a numeric index. I think this is a niche feature and of little to no use to end users.


sauerbraten:
Builds using Go versions 1.12 or earlier fail because we use https://golang.org/pkg/reflect/#Value.IsZero now, which was added in 1.13.

sauerbraten:
AppVeyor doesn't have Go 1.14 yet: https://github.com/appveyor/ci/issues/3325

sauerbraten:
Strangely, the Travic build results don't show up here anymore, but they passed: https://travis-ci.org/github/CloudyKit/jet/builds/676118992

tooolbox:
Interesting.  Seems like these are improvements without much potential for negative impact.  Are you thinking v3.1.0?

> This PR removes support for indexing into structs with a numeric index. I think this is a niche feature and of little to no use to end users.

I agree, this is an unusual one.

sauerbraten:
Technically, since we remove a feature, it would have to be v4.0.0... 🤔
But I still doubt anyone's using it, so I'd be fine with v3.1.0, yeah.

PR #151, opened 2020-06-22
Try & Recover

sauerbraten:
This pull request introduces the optional keyword 'atomic' at the end of an include statement. Marking an include statement as atomic means that all panics (= runtime errors) occuring during execution of the named template are swallowed. The output created by the named template is included in the output of the including template only if no panic occured, and left out entirely if any panic was swallowed.

The tests explain the usage of 'atomic' pretty well I think.

This feature would help our usage of jet, where we have certain templates that depend on third-party services/APIs through custom helper functions, and rather than having the entire page rendering fail we would like to simply omit a section of the page.

Still unsolved is the problem of logging: we still want to be notified of the runtime error somehow. Either we could make jet stream these non-critical errors via a channel, or we collect them during execution and return a certain "swallowed errors" collection type from Execute() that users of jet can check for and act on. The second option would mean the API only changes in semantics, and users that opt into atomic includes can update their Execute() calls accordingly at the same time.


sauerbraten:
This is based on the feature branch for https://github.com/CloudyKit/jet/pull/150, so that should be merged first!

tooolbox:
> Marking an include statement as atomic means that all panics (= runtime errors) occuring during execution of the named template are swallowed. The output created by the named template is included in the output of the including template only if no panic occured, and left out entirely if any panic was swallowed.

While the feature seems handy for the use case you mention, I feel like there's a better word than "atomic" available, since atomic usually means concurrent-safe.

It's really about error handling, so maybe "soft", "safe", "optional"?  Or "recover", since you're saying to catch and ignore the panic within the include rather than bubbling it up further.

Actually, there's already an analogous call in Jet: [`includeIfExists`](https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax#includeifexists) which handles the case of a missing template.  How about `includeSafe`?

Could even set it up to work the same way, so you can optionally add specific output if the inclusion fails.

```handlebars
{{if ok := includeSafe("myFailableTemplate"); !ok}}
    <p>The template failed to execute.</p>
{{end}}
```

Actually, if you return the error, this could handle your requirement of feeding errors back to the rest of your code:
```handlebars
{{if err := includeSafe("myFailableTemplate"); err != nil}}
    {{ReportError(err)}}
{{end}}
```

👍

sauerbraten:
I really like the idea of calling it "recover" or "recovering" instead of "atomic"! 👌 I originally meant "atomic" in regards to the output, which is all-or-nothing, but I agree it's not a very clear name.

I'm not entirely sold on implementing it as a function like `includeIfExists` is. In that case, it's impossible to ensure the template exists at parsing instead of run time. Instead, I would like to add the "recovering" behavior to `includeIfExists`, for example by introducing an additional optional argument, or a built-in wrapping function that swallows any pancis of inner function calls (`{{ ignoringPanics(includeIfExists("./bla.jet")) }}`).

tooolbox:
> I really like the idea of calling it "recover" or "recovering" instead of "atomic"! 👌 I originally meant "atomic" in regards to the output, which is all-or-nothing, but I agree it's not a very clear name.

Great :)

> I'm not entirely sold on implementing it as a function like includeIfExists is. In that case, it's impossible to ensure the template exists at parsing instead of run time.

Interesting, I hadn't thought of that.

I was actually more approaching it from the angle of "modify the 'include' operator rather than adding a new param".  So even if it's not a function call, it would be `{{includeSafe "template" .}}` or some such.  I just think from an API design perspective that's more consistent, to have `includeX` wherever we have a form of `include` with special behavior, whether or not it's a function call.

> add the "recovering" behavior to includeIfExists, for example by introducing an additional optional argument, or a built-in wrapping function that swallows any panics of inner function calls `{{ ignoringPanics(includeIfExists("./bla.jet")) }}`

Hm, can't say I'm in love with that nested approach.  How about `includeSafeIfExists()`?  Kinda wordy but better in my mind.  `includeSafe` and `includeSafeIfExists()` would together handle the scenario of recovering from a problematic template, with the option to be sure it's there at parse time or not--would that suffice?

If you want to add the broad capability of recovering from panics anywhere, perhaps a builtin might be most appropriate?

```handlebars
{{recover err}}
    {{SaveError(err)}}
    {{yield errorBlock(err)}}
{{with}}
    {{include "mytemplate" .}}
    {{DangerousFunction()}}
{{end}}
```

That's new syntax, and maybe there's a better way to do it, but might be more flexible and readable than a nested function call.

Could also do `{{if err := recover; err != nil}}{{with}}{{end}}` or `{{range err := recover}}{{else}}{{end}}` or use `recover()` or use blocks somehow.

(I'm not even sure that general error/panic handling is necessary if we had both `includeSafe` and `includeSafeIfExists()` in the toolbox, because at that point I think 90% of use cases are covered, but I could be wrong.)

I know I'm fully armchair quarterbacking at this point, since I have no idea what it would take to do the above and you're the one doing the hard work of implementation, but I'm just trying to give some feedback of what I as a Jet user would want to see in the API, what feels clean, etc.

Thanks!

sauerbraten:
> Hm, can't say I'm in love with that nested approach.

I just realized that it wouldn't even work, since of course the inner function runs before the outer one, so I can't recover panics from the inner one.

> If you want to add the broad capability of recovering from panics anywhere, perhaps a builtin might be most appropriate?

Really like this idea, actually! That way, include, includeIfExists, block yielding and everything else is covered! 👍

I'd go for this, though, which I hope will feel more intuitive to most users:
```handlebars
{{try}}
  {{include "./foo.jet"}}
{{recover err}}
  {{log(err)}}
{{end}}
```

sauerbraten:
Closed in favor of a fresh start in #152.

tooolbox:
> Really like this idea, actually! That way, include, includeIfExists, block yielding and everything else is covered! 👍

Yup 😄 

> I'd go for this, though, which I hope will feel more intuitive to most users:

Ha, very nice!  I think that's perfect, a lot more familiar.

tooolbox:
Actually, it occurs to me, would `try`/`catch` be better because it's more familiar?  I know `recover` is more Go-style, but we also don't have `try`.  Anyway, I'm on the fence about that one, I think I lean towards `try`/`catch` but both are good, so whatever you decide.

Thanks for doing this!

PR #152, opened 2020-06-23
Introduce try/catch control structure

sauerbraten:
`{{try}} .. {{end}}` or `{{try}} .. {{catch [err]}} .. {{end}}` executes a list of statements, recovering from any panic (runtime errors) inside. When using try with catch, the statement list inside catch/end is executed in the event of a recovery.

catch takes an optional identifier name under which the error will be available inside the catch statement list.

In case of no runtime error, the output generated by the try statement list will be kept. In case of any panic, no output from the try statement list will be kept.

Born out of the ashes of #151.


tooolbox:
Brilliant!  (I didn't deep-dive into the internals to review what you did but the concept is great.)

Do you think it will affect performance much?  Probably only if you're dealing with tons of panics, I imagine.

sauerbraten:
> Do you think it will affect performance much?

Well it only impacts performance when you actually use it. If you use it, there's certainly a hit to the speed of outputting rendered results and in increase in memory consumption, because we buffer output until the entire block was executed. Other than that, it shouldn't have a big impact, it only shuffles around a few variables before and after executing the statement list. Deferred functions don't have any significant overhead in recent Go versions, either, so the panic handling should be ok, too.

sauerbraten:
I liked the suggestion to rename "recover" to "catch" you had here: https://github.com/CloudyKit/jet/pull/151#issuecomment-648292731. Coming up!

tooolbox:
> Well it only impacts performance when you actually use it. If you use it, there's certainly a hit to the speed of outputting rendered results and an increase in memory consumption, because we buffer output until the entire block was executed. Other than that, it shouldn't have a big impact, it only shuffles around a few variables before and after executing the statement list. Deferred functions don't have any significant overhead in recent Go versions, either, so the panic handling should be ok, too.

That makes sense.  Yeah, it's something you can use if you need it.

> I liked the suggestion to rename "recover" to "catch" you had here: #151 (comment). Coming up!

👍   I think this is great, a good tool for the arsenal :)

tooolbox:
Sweet!  🎉 

Are you going to tag v3.0.1?  :)

sauerbraten:
Very likely will tag v4.0.0 soon, there are breaking API changes once again... 🤷‍♂️

tooolbox:
Huh, kk, I mean this is not breaking right, it's additive?  Would it make sense to tag it before you do the major version jump?

sauerbraten:
A change to runtime.Set() is kinda breaking, see
- https://github.com/CloudyKit/jet/pull/152/commits/42118a3074341faf243ff22723b1fa22bddb6d9d
- https://github.com/CloudyKit/jet/pull/152/commits/42ce0ed64806098dc0eac404860dadf011a16066
- https://github.com/CloudyKit/jet/commit/af8a04ba4e8481a202aece2ef3bdf5a602a1b44d

tooolbox:
Huh, hadn't noticed that earlier.  Yeah, makes sense.

PR #161, opened 2020-07-03
Update module path

razonyang:


sauerbraten:
Oh, thanks a lot! Completely missed that!

PR #169, opened 2020-07-22
add support for function capturing syntax

### LINKED ISSUE #162: Add function capturing syntax
From https://github.com/CloudyKit/jet/issues/82#issuecomment-653627407:

> would be kinda cool to do something like [Gleam's function capture](https://gleam.run/tour/functions.html#function-capturing) where you can specify what position you want the argument in, so like `{{ foo()|bar(_, param2, param3)|baz(param1, _, param3) }}` (Gleam probably took that from somewhere, but that's just where I've seen it recently.)

cc @tooolbox

sauerbraten:
closes #162


sauerbraten:
cc @tooolbox

tooolbox:
Looks incredible, thanks for the great work 😃

PR #173, opened 2020-09-01
Trim Whitespace

### LINKED ISSUE #172: Whitespace Removal
Just a thought, would like feedback from @sauerbraten whenever possible 😉 

One thing that Go Templates has over Jet is the ability to [trim preceding and trailing whitespace](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) using `{{- ` and ` -}}`.  This comes in handy sometimes and would be cool to have in Jet (not necessarily with that exact syntax, but I haven't thought hard about a better alternative).

Since at this point you're possibly the most familiar with Jet's inner workings, I'm interested in your take on how feasible this would be to implement?

tooolbox:
Closes #172 

...That was more straightforward than I thought 😉 

Depending on certain factors, I might also like to backport this to v3 or v4, but this is a start.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -47,7 +47,29 @@
   - [extends](#extends)
   - [import](#import)
 
-## Comments
+## Delimiters
+
+Template delimiters are `{{` and `}}`.  Delimiters can use `.` to output the execution context, and they can also contain many different literals, expressions, and declarations.
+
+    hello {{ . }} <!-- with context of "world", outputs "hello world" -->
+
+Jet can also be configured to use alternative delimiters, such as `[[` and `]]`.
+
+    hello [[ . ]] <!-- equivalent to the above -->
+
+### Whitespace Trimming
+
+By default, all text outside of template delimiters is copied verbatim when the template is parsed and executed, including whitespace.
+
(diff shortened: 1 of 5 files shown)


sauerbraten:
Very cool, thanks for taking the time!

tooolbox:
> Very cool, thanks for taking the time!

Yeah, my curiosity was piqued and it was halfway done before I realized what I was doing 😉 

> Do you want to add a "delimiters" section to the syntax reference describing this feature?

Sure, done.  Feel free to tweak what I wrote if the urge comes over you.

...As a note, do you prefer no syntax highlighting?  Your docs may be an improvement over [the wiki](https://github.com/CloudyKit/jet/wiki/3.-Jet-template-syntax) but I liked the color.

sauerbraten:
The documentation is great, thanks!

> As a note, do you prefer no syntax highlighting?

I guess I started to use intend markdown syntax for my code blocks because it was easier to read in the editor. For these short snippets, I guess I didn't miss syntax highlighting, but I'm certainly not against it. If you want to bring it back in a PR, I'll make sure we always use it in the future! 👍

Issue #181, opened 2020-10-26
Thread safe jet.Template


Dynom:
In line of my previous Issue https://github.com/CloudyKit/jet/issues/180 I've considered of re-using instances of `jet.Template` as much as possible. However I'm pretty sure that in their current form they are not safe for concurrent use:
```go
func (t *Template) ExecuteI18N(translator Translator, w io.Writer, variables VarMap, data interface{}) (err error) {
	st := pool_State.Get().(*Runtime)
	defer st.recover(&err)

	st.blocks = t.processedBlocks // <-- Unsafely copying a Map
	st.translator = translator
	st.variables = variables
	st.set = t.set                // <-- Unsafely referencing *Set, copying over at least Set.globals and Set.escapee
	st.Writer = w

	// resolve extended template
	for t.extends != nil {
		t = t.extends         // <-- Unsafely mutating `t`
	}

	if data != nil {
		st.context = reflect.ValueOf(data)
	}

	st.executeList(t.Root)        // <-- st.executeList() unsafely accesses t.Root.Nodes
	return
}
```

A test to show data races, when run as: `go test -race -test.run ^TestJetConcurrency$`
```go
// jetversion5 "github.com/CloudyKit/jet/v5"
func TestJetConcurrency(t *testing.T) {
	views := jetversion5.NewHTMLSet(".")
	tpl, err := views.LoadTemplate("foo", "{{if true}}Hi {{- .Name -}}!{{end}}")
	if err != nil {
		t.Errorf("Error in test %s", err)
	}

	for i := 0; i < 3; i++ {
		t.Run(fmt.Sprintf("CC_%d", i), func(t *testing.T) {
			t.Parallel()

			err = tpl.Execute(ioutil.Discard, nil, struct{ Name string }{Name: "John"})
			if err != nil {
				t.Errorf("Error in test %s", err)
			}
		})
	}
}
```

Version: v5.1.1

sauerbraten:
- `st.blocks = t.processedBlocks` is okay as long as the template is not re-parsed while it is executed, which I'm pretty sure is impossible from outside the jet package and isn't done inside the package.
- `st.set = t.set` only copies a pointer to the Set, nothing inside it
- `t = t.extends` also only sets a local copy of the `tpl` pointer inside your t.Run() to a new value
- `st.executeList(t.Root)` is (like accessing `t.processedBlock`) okay as long as the template is not re-parsed during execution

[...]

PR #182, opened 2020-10-26
cache more aggressively; remove Set.LoadTemplate()

### LINKED ISSUE #180: Performance: Steep decline between v2 and v5
I've been doing several proof of concepts and the performance of v5's loading of templates is significantly less than of the v2 branch:
```
BenchmarkJets
BenchmarkJets/Jetv2
BenchmarkJets/Jetv2-8         	   48260	     24372 ns/op	    5040 B/op	      87 allocs/op
BenchmarkJets/Jetv5
BenchmarkJets/Jetv5-8         	    3828	    267862 ns/op	   36687 B/op	     456 allocs/op
```

## The setup
[Versions](https://github.com/CloudyKit/jet/releases):
- github.com/CloudyKit/jet v2.1.2+incompatible
- github.com/CloudyKit/jet/v5 v5.1.1

The implementation and templates are identical, with the exception of some differences required for the different Jet versions:
```sh
$ find . -type f -name "*.jet" -print -exec diff -Naruw ../jetv2/{} ../jetv5/{} \;
```
```diff
./views/leaderboard.jet
--- ../jetv2/./views/leaderboard.jet	2020-10-20 10:18:04.000000000 +0200
+++ ../jetv5/./views/leaderboard.jet	2020-10-20 12:55:05.000000000 +0200
@@ -1,8 +1,8 @@
-{{import "partials/leaderboard_row.jet"}}
+{{import "/partials/leaderboard_row.jet"}}

 <table>
     <!-- Control flow is a must. This includes iterating over arrays, if/elseif/else statements and ternaries. -->
-    {{range competitor := .Leaderboard}}
+    {{range _, competitor := .Leaderboard}}
     <!-- comments do not appear in the rendered HTML -->
         {* `$competitor` is of type `User` *}
         {{yield leaderboard_row(user=competitor)}}
./views/index.jet
--- ../jetv2/./views/index.jet	2020-10-20 10:18:04.000000000 +0200
+++ ../jetv5/./views/index.jet	2020-10-20 12:45:10.000000000 +0200
@@ -1,6 +1,6 @@
-{{extends "layouts/main.jet"}}
+{{extends "/layouts/main.jet"}}

-{{import "partials/ui.jet"}}
+{{import "/partials/ui.jet"}}

 {{block body()}}
     <script type="text/javascript">
./views/layouts/main.jet
./views/layouts/header.jet
--- ../jetv2/./views/layouts/header.jet	2020-10-26 09:09:56.000000000 +0100
+++ ../jetv5/./views/layouts/header.jet	2020-10-20 12:45:03.000000000 +0200
@@ -1,4 +1,4 @@
-{{import "partials/ui.jet"}}
+{{import "/partials/ui.jet"}}

 <div id="header">
     <div id="menu">
./views/layouts/footer.jet
./views/partials/ui.jet
./views/partials/leaderboard_row.jet
```
The files:
```
.
├── jet.go
└── views
    ├── index.jet
    ├── layouts
    │   ├── footer.jet
    │   ├── header.jet
    │   └── main.jet
    ├── leaderboard.jet
    └── partials
        ├── leaderboard_row.jet
        └── ui.jet
```

## The benchmark
```go
import (
	jetversion2 "github.com/CloudyKit/jet"
	jetversion5 "github.com/CloudyKit/jet/v5"
	// ..
)

func BenchmarkJets(b *testing.B) {
	if wd, err := os.Getwd(); err == nil && filepath.Base(wd) == "template" {
		_ = os.Chdir("..")
	}

	tr := reflect.ValueOf(template.TranslateFunc("en"))
	data := template.ExampleContext()

	b.Run("Jetv2", func(b *testing.B) {

		views := jetversion2.NewHTMLSet("./template/jetv2/views")
		views.SetDevelopmentMode(false)

		for i := 0; i < b.N; i++ {
			t, err := views.GetTemplate("index.jet")
			if err != nil {
				b.Errorf("Error in benchmark %s", err)
			}

			err = t.Execute(ioutil.Discard, jetversion2.VarMap{"T": tr}, data)
			if err != nil {
				b.Errorf("Error in benchmark %s", err)
			}
		}
	})

	b.Run("Jetv5", func(b *testing.B) {

		views := jetversion5.NewHTMLSet("./template/jetv5/views")
		views.SetDevelopmentMode(false)
		views.SetExtensions([]string{""}) // v5 specific optimisation

		for i := 0; i < b.N; i++ {
			t, err := views.GetTemplate("index.jet")
			if err != nil {
				b.Errorf("Error in benchmark %s", err)
			}

			err = t.Execute(ioutil.Discard, jetversion5.VarMap{"T": tr}, data)
			if err != nil {
				b.Errorf("Error in benchmark %s", err)
			}
		}
	})
}
```

## Considerations 
I noticed quite a bit of time spent in io (`Loader.Exists`). I wrote a quick-n-dirty Loader that pre-fetches all *.jet files and adds them to a map for faster subsequent lookups. This reduces the execution time to 17 ms (down from 267ms). But I think it only masks the problem, instead of properly solving it.

I noticed many invocations to `Loader.Exists` and `Loader.Load()` for already observed files. My *guess* so far is that some state should be preserved in `func (st *Runtime) executeList` when `st.executeInclude()` is invoked. Or that the way file paths are canonicalised should be more in line with how templates can be used in version 5. e.g.:
- `{{ include "./index.jet" }}`
- `{{ include "/index.jet" }}`
- `{{ include "index.jet" }}`

might all point to the same file.

Screenshot 2020-10-26 at 16 38 55
Screenshot 2020-10-26 at 16 40 21

## Further findings
After I've plugged a different Loader I managed to get it down to the following. However I think this isn't the approach to take. I think the lazy-loading approach is probably a fine route to take, but right now it appears to not resolving efficiently.
```
BenchmarkJets
BenchmarkJets/Jetv2
BenchmarkJets/Jetv2-8         	   49384	     23685 ns/op	    5041 B/op	      87 allocs/op
BenchmarkJets/Jetv5
BenchmarkJets/Jetv5-8         	   69686	     17440 ns/op	    5091 B/op	      72 allocs/op
```

sauerbraten:
~~Keeping track of wether or not to cache templates that are imported/extended during Parse() is too much work, so Parse() is gone for now.~~ LoadTemplate() is removed in favor of a new in-memory Loader where you can add templates on-the-fly (which is also used for tests without template files). Using it together with a file loader via a multi loader restores the previous functionality.

I believe this fixes #180.

### DIFF
--- a/loader.go
+++ b/loader.go
@@ -15,17 +15,21 @@
 package jet
 
 import (
+	"bytes"
+	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 )
 
 // Loader is a minimal interface required for loading templates.
 type Loader interface {
 	// Exists checks for template existence.
-	Exists(path string) (string, bool)
+	Exists(templatePath string) (string, bool)
 	// Open opens the underlying reader with template content.
-	Open(path string) (io.ReadCloser, error)
+	Open(templatePath string) (io.ReadCloser, error)
 }
 
 // OSFileSystemLoader implements Loader interface using OS file system (os.File).
@@ -39,22 +43,58 @@ var _ Loader = (*OSFileSystemLoader)(nil)
 // NewOSFileSystemLoader returns an initialized OSFileSystemLoader.
 func NewOSFileSystemLoader(dirPath string) *OSFileSystemLoader {
 	return &OSFileSystemLoader{
-		dir: dirPath,
+		dir: filepath.FromSlash(dirPath),
 	}
 }
 
 // Open opens a file from OS file system.
-func (l *OSFileSystemLoader) Open(path string) (io.ReadCloser, error) {
-	return os.Open(filepath.Join(l.dir, path))
+func (l *OSFileSystemLoader) Open(templatePath string) (io.ReadCloser, error) {
+	return os.Open(filepath.Join(l.dir, filepath.FromSlash(templatePath)))
 }
 
 // Exists checks if the template name exists by walking the list of template paths
 // returns true if the template file was found
-func (l *OSFileSystemLoader) Exists(path string) (string, bool) {
-	path = filepath.Join(l.dir, path)
-	stat, err := os.Stat(path)
+func (l *OSFileSystemLoader) Exists(templatePath string) (string, bool) {
+	templatePath = filepath.Join(l.dir, filepath.FromSlash(templatePath))
+	stat, err := os.Stat(templatePath)
 	if err == nil && !stat.IsDir() {
-		return path, true
+		return templatePath, true
 	}
 	return "", false
+}
+
+type InMemLoader struct {
+	files map[string][]byte
+}
+
+// compile time check that we implement Loader
+var _ Loader = (*InMemLoader)(nil)
+
+func NewInMemLoader() *InMemLoader {
+	return &InMemLoader{
+		files: map[string][]byte{},
+	}
+}
+
+func (l *InMemLoader) Open(templatePath string) (io.ReadCloser, error) {
+	f, ok := l.files[templatePath]
+	if !ok {
+		return nil, fmt.Errorf("%s does not exist", templatePath)
+	}
+
+	return ioutil.NopCloser(bytes.NewReader(f)), nil
+}
+
+func (l *InMemLoader) Exists(templatePath string) (string, bool) {
+	_, ok := l.files[templatePath]
+	if !ok {
+		return "", false
+	}
+	return templatePath, true
+}
+
+func (l *InMemLoader) Set(templatePath, contents string) {
+	templatePath = filepath.ToSlash(templatePath)
+	templatePath = path.Join("/", templatePath)
+	l.files[templatePath] = []byte(contents)
 }
--- a/template.go
+++ b/template.go
@@ -19,6 +19,7 @@
 package jet
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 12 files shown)


loader.go
```diff
@@ -39,22 +42,56 @@ var _ Loader = (*OSFileSystemLoader)(nil)
...
+type InMemLoader struct {
+	files map[string][]byte
+}
+
+// compile time check that we implement Loader
+var _ Loader = (*InMemLoader)(nil)
+
+func NewInMemLoader() *InMemLoader {
+	return &InMemLoader{
+		files: map[string][]byte{},
```
Dynom, line 74: Accessing `l.files` without mutex allows for race conditions. There a few areas within Jet that could use a second look (see also: https://github.com/CloudyKit/jet/issues/181)

template.go
```diff
@@ -174,24 +147,33 @@ func (s *Set) getTemplate(templatePath string) (t *Template, err error) {
 		}
 	}
 
-	return s.getTemplateFromLoader(templatePath)
+	t, err = s.getTemplateFromLoader(templatePath)
+	if err == nil && !s.developmentMode {
+		s.templates[templatePath] = t
```
Dynom, line 152: Access to `s.templates` should either be guarded with a mutex or use the type [`sync.Map`](https://golang.org/pkg/sync/#Map), as I suspect this type is likely to be a high read on single write.

sauerbraten:
Thanks for your comments, @Dynom! I'd like to address Jet's concurrency-safety later in a new PR targeting #181, and ignore these for now, okay?

loader.go
sauerbraten: will be addressed in fix for #181

template.go
sauerbraten: will be addressed in fix for #181

sauerbraten:
I re-added the Set.Parse() method and cleaned up the commit history.

PR #183, opened 2020-10-28
add Cache, simplify Loader, revamp NewSet

### LINKED ISSUE #179: Add time-based template caching capability
The use case: I want to be able to replace the templates on-the-fly but with reasonable termplate rendering performance. I know that there is 'development mode' flag already, but obviously I'm not going to use this on high-traffic production environment.

Proposal: Introduce time-based caching. With this I can have both high-performance rendering performance and the ability to replace the templates on-the-fly. The cache item expiration time should be configurable.

I have some ideas already on the implementation, but I'd like to hear your feedback first on the idea.

sauerbraten:
- Set type and related code are moved out of template.go into set.go
- template.go -> exec.go
- introduce Cache interface to allow custom caching strategies/behaviors/mechanisms (closes #179)
- added better doc comments to Loader interface
- remove path from Exists() return values
- normalize input paths for all methods in InMemLoader
- Set's constructors and SetX() methods are unified into NewSet() with sane defaults and option functions
- concurrency tests are added for Template.Execute() and Set.GetTemplate() (addresses #181)

Altogether, this would mean a v6 release with major breaking changes in Jet's API.


sauerbraten:
@tooolbox: pinging you because this contains a change you previously didn't like: Loader.Exists() doesn't return the actual path anymore.

I introduced this because Jet builds canonical paths for each referenced template now before looking them up in the loader, i.e. no matter what path is used to reference a certain template from somewhere else, the lookup to the loader will always be an absolute, slash-delimited path. I don't see the point of returning that same path. Jet doesn't need it from any Loader, and it makes sense for no Loader implementations provided by the package. Outside implementations are free to have a Normalize() method if they deal with a second path scheme apart from Jet's.

In general, feedback on this PR is greatly appreciated if you can find the time!

Issue #110, opened 2019-03-05
Option to not abourt Template rendering


kabukky:

mamarx:
+1

tooolbox:
I believe this is solved with https://github.com/CloudyKit/jet/pull/152

sauerbraten:
Mhmm... not really I'd say. This issue asks to silently ignore undefined variables etc by default (or at least by a configurable flag). Try/catch doesn't do that.

tooolbox:
Gotcha.  Well, it seemed like a way to avoid aborting the entire template process when you find an invalid expression.  In either case, dunno that it's a great idea to ignore undefined variables; seems like it would make things more buggy.

🤷‍♂️

tonyhb:
For user defined templates this is a pretty important feature.  You can't guarantee that user's input (either templates or variables) are always accurate.  I think an option to use `Resolve()` vs `resolve()` and ignore errors makes a lot of sense!

sauerbraten:
Jet isn't intended to be used with unchecked user input, at least it's definitely not something I would recommend.

The Resolve() vs resolve() proposition sounds interesting, even though the code would fall apart the first time you pass a buggy resolved value to a function or something...

tonyhb:
I hear you.  RE. unchecked user input, was actually attempting to compile this with GOOS=js and GOARCH=wasm to run this in a separate webassembly environment.  I'll investigate other things for now :)

tonyhb:
For the fun of it, I spent 20 minutes hacking together a feature that allows "custom resolvers".  The diff is here:

https://github.com/CloudyKit/jet/compare/master...tonyhb:feature/custom-resolver

This lets us do whatever we want, including skipping errors, getting a list of all variables used in a template, etc.

It adds a single interface - `Resovler`:
```go
type Resolver interface {
	Resolve(name string) (reflect.Value, error)
}
```

This allows us to define custom functions overriding default functionality in Resolve.  You can always get the default resolver functionality within Runtime by calling `runtime.DefaultResolver()`.

Usage:

```go
type CustomResolver struct {
  default jet.Resolver
}

func (cr *CustomResolver) Resolve(name string) (reflect.Value, error) {
        fmt.Printf("resolving %s\n", name)
        return cr.default(name)
}

var buf bytes.Buffer
tpl, err := jet.NewSet(jet.NewInMemLoader()).Parse("some-tpl.tpl". source)
runtime := tpl.Runtime()
[...]

PR #186, opened 2020-12-22
Custom resolvers

tonyhb:
Opening a PR with a _prototype_ as a means to discuss custom resolvers.  [Based off of this comment](https://github.com/CloudyKit/jet/issues/110#issuecomment-749494289) in #110, custom resolvers allow us to override default functionality when executing a template.  This would open the possibility to:

- Implement logging and tracing
- Implement custom error handling logic
- Implement introspection without navigating the AST

This PR is purely a concept.  It has no tests, and it exposes *Runtime directly from a template, which I do not like.  So, word of warning, this PR is _not ready_ and the aim here is to open a discussion with maintainers and owners regarding whether Resolver interfaces can be considered.

Stealing from my previous comment:

It adds a single interface - `Resovler`:
```go
type Resolver interface {
	Resolve(name string) (reflect.Value, error)
}
```

This allows us to define custom functions overriding default functionality in Resolve.  You can always get the default resolver functionality within Runtime by calling `runtime.DefaultResolver()`.

Usage:

```go
type CustomResolver struct {
  default jet.Resolver
}

func (cr *CustomResolver) Resolve(name string) (reflect.Value, error) {
        fmt.Printf("resolving %s\n", name)
        // Here, we can ignore the error from our default resolver, choose to do no resolving at all, etc.
        return cr.default(name)
}

var buf bytes.Buffer
tpl, err := jet.NewSet(jet.NewInMemLoader()).Parse("some-tpl.tpl". source)
runtime := tpl.Runtime()
runtime.WithResolver(&customResolver{})
tpl.ExecuteWith(runtime, buf, nil, nil)
```

I don't like how I have to expose the `Runtime()` func within templates to get the default resolver at all to be honest, but yeah, like I said - it's a hacky POC to show that interfaces here might help.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -40,9 +40,12 @@ func (scope VarMap) SetWriter(name string, v SafeWriter) VarMap {
 	return scope
 }
 
-func (t *Template) ExecuteWith(r Resolver, w io.Writer, variables VarMap, data interface{}) (err error) {
+func (t *Template) Runtime() *Runtime {
 	st := pool_State.Get().(*Runtime)
-	st.WithResolver(r)
+	return st
+}
+
+func (t *Template) ExecuteWith(st *Runtime, w io.Writer, variables VarMap, data interface{}) (err error) {
 	return t.execute(st, w, variables, data)
 }
 
--- a/exec.go
+++ b/exec.go
@@ -40,11 +40,21 @@ func (scope VarMap) SetWriter(name string, v SafeWriter) VarMap {
 	return scope
 }
 
+func (t *Template) ExecuteWith(r Resolver, w io.Writer, variables VarMap, data interface{}) (err error) {
+	st := pool_State.Get().(*Runtime)
+	st.WithResolver(r)
+	return t.execute(st, w, variables, data)
+}
+
 // Execute executes the template into w.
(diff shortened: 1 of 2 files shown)


tonyhb:
Looks like there are some issues with templates halting execution when eg. iterating over nil ranges.  So, no matter what, it doesn't seem like adding a Resolver interface magically lets us customize execution.  Kinda writing off this idea but will keep this open for discussion.

exec.go
```diff
@@ -40,11 +40,24 @@ func (scope VarMap) SetWriter(name string, v SafeWriter) VarMap {
 	return scope
 }
 
+func (t *Template) Runtime() *Runtime {
+	st := pool_State.Get().(*Runtime)
```
sauerbraten, line 44: This will give you *any* runtime from the pool, possibly a new one, not the one executing the template.

sauerbraten: Instead, you should create a runtime, then create a custom resolver to which you pass the runtime, then set the resolver in the runtime, then execute the template.

sauerbraten:
I like the overall idea, however I believe it would be good to give users a way to hook and redirect variable declarations as well in order to make a custom resolver really useful. I'm picturing something like a user defined and controlled `reflect.Value` store.

Issue #187, opened 2021-01-03
using custom delimiters breaks recognition of comments


jan-herout:
After tinkering with Jet using custom delimiters I was able to reproduce the following problem directly on Jet test suite. 

- modified file : `testData/custom_delimiters.jet`
- first line now reads: `{* comment *}[[ . ]]`
- test fails

```
c:\GitHub\jet>go test
--- FAIL: TestParseTemplateWithCustomDelimiters (0.00s)
    parse_test.go:46: Unexpected tree on custom_delimiters.jet Got:
        {* comment *}{{.}}
        {{singleValue}}
        {{nil}}
        {{""}}
        {{val.Field}}
        {{url("")}}
        {{url("", "") | pipe}}
        {{url("") | pipe | pipe}}
        {{url("", "").Field | pipe}}
        {{url("", "").Method("") | pipe}}
        Expected:
        {{.}}
        {{singleValue}}
        {{nil}}
        {{""}}
        {{val.Field}}
        {{url("")}}
        {{url("", "") | pipe}}
        {{url("") | pipe | pipe}}
        {{url("", "").Field | pipe}}
        {{url("", "").Method("") | pipe}}
FAIL
exit status 1
FAIL    github.com/CloudyKit/jet/v6     1.775s
```

Might it be that this is the source of the problem? See below... It seems that methow `WithDelims` should accept full delimiter set to allow for correct parsing of comments.

```
// state functions
func lexText(l *lexer) stateFn {
	for {
		// look for first character of the custom delimiter. if not found, break
		// this EXCLUDES comment defined as {* from the parse, and sends it on output as text ???
		if i := strings.IndexByte(l.input[l.pos:], l.leftDelim[0]); i == -1 {
			l.pos = Pos(len(l.input))
			break
		} else {
			l.pos += Pos(i)
			if strings.HasPrefix(l.input[l.pos:], l.leftDelim) {
				ld := Pos(len(l.leftDelim))
				trimLength := Pos(0)
				if strings.HasPrefix(l.input[l.pos+ld:], leftTrimMarker) {
					trimLength = rightTrimLength(l.input[l.start:l.pos])
				}
				l.pos -= trimLength
				if l.pos > l.start {
					l.emit(itemText)
				}
				l.pos += trimLength
				l.ignore()
				return lexLeftDelim
			}
			// with custom delimiters, this will NEVER be activated
			if strings.HasPrefix(l.input[l.pos:], leftComment) {
				if l.pos > l.start {
[...]

Issue #111, opened 2019-03-13
Dumping Vars in templates for development


Drudwyn:
For rapid cross-team development, it would be useful to allow a developer to dump Vars in a template in a pretty print, such as Go-Spew. Is this achievable?

ryanlath:
Not pretty, but I use this:
```
views.AddGlobalFunc("dump", func(a jet.Arguments) reflect.Value {
	a.RequireNumOfArguments("dump", 1, 1)
	return reflect.ValueOf(fmt.Sprintf("%#v", a.Get(0)))
})
```

Then in template:
`{{ dump(VAR) }}`

ryanlath:
I should really read the docs.  There's already a built in func:
`{{ json(VAR) }}`
Then, I'm sure there's a golang JSON prettifier out there.

Dexus:
Would be nice to have a debug function, that returns all variables that are availabel in the context.

Don't like to use some variables that not there to read the error with the varaibels.

annismckenzie:
Could you take a crack at that? It sure would be nice.

sauerbraten:
I'd like dump() to be smart and versatile, so I imagine an API that allows the following calls:
```
dump()
dump(2)
dump("foo")
dump("foo", "bar", "lorem", "ipsum")
```

In the first case, I'd expect dump to print multiple sections:
1. everything in Runtime.context
2. everything in Runtime.variables
3. everything in Runtime.set.globals
3. everything in Runtime.blocks

In the second case, when called with a single int argument, dump should interpret that as the maximum depth of parent scopes to walk up, and additionally print Runtime.parent.variables and Runtime.parent.parent.variables between the Runtime.variables section and the Runtime.set.globals section. (This means the first case is just like the second case, just with an implied depth of 0.)

The third and fourth case are handled the same way: for a call with one or more string arguments, dump would treat each string as an identifier name and print the name, type, and value (and maybe even scope?) of whatever was found when looking up each identifier. The lookup depth in parent scopes is unlimited, like in normal jet usage. For example,
```
foo [string]:
  "bar"
```
or
```
foo(type="text", label, id, value="", required=false) [block]:
  <div class="form-field">
      <label for="{{ id }}">{{ label }}</label>
      <input type="{{ type }}" value="{{ value }}" id="{{ id }}" {{ required ? "required" : "" }} />
  </div>
```

So an identifier could be describe a variable, something from the context, a global, or a block. (Maybe it could even be a template path and dump the template AST?)

[...]

PR #189, opened 2021-02-01
dump() built-in function (#111)

jan-herout:
The following is - perhaps naive? - attempt to satisfy requirement described in issue #187. Let me know what you think - Jet internals are mystery to me, at least for now.

To be honest, I am not sure why comparison against master contains also changes implemented by 4beb91506060bd8900ec0f7d42e401d21300350f and f67928da339229e4ccb9b198a6428b8f077cf1c9 - that is confusing.

Regards, 

Jan


jan-herout:
Hmm, not sure why this fails, it has nothing to do with the change.

```
FAIL
FAIL	github.com/CloudyKit/jet/v6	0.072s
?   	github.com/CloudyKit/jet/v6/examples/asset_packaging	[no test files]
?   	github.com/CloudyKit/jet/v6/examples/asset_packaging/assets/templates	[no test files]
?   	github.com/CloudyKit/jet/v6/examples/todos	[no test files]
?   	github.com/CloudyKit/jet/v6/jettest	[no test files]
```

jan-herout:
I have read the ticket second time, and my attempt is something different then what was asked for. Do not waste your time, I might put it here in the end. I will sleep on it.

https://gitlab.com/Herout/jetfunc

jan-herout:
OK, I had to try again. Please let me know, what do you think - could this be the solution to #187? Thanks!

Regards,

Jan.

default.go
```diff
@@ -151,6 +153,20 @@ func init() {
...
 		})),
+		"dump": reflect.ValueOf(Func(func(a Arguments) (result reflect.Value) {
+			runtime := a.runtime
+			vars := runtime.variables
+			var sb strings.Builder
+			sb.WriteString("--dump start\n")
+			for k, v := range vars {
+				sb.WriteString(k)
+				sb.WriteString("=")
+				fastprinter.PrintValue(&sb, v)
```
sauerbraten, line 164: For a debugging function like dump(), I'd say it's fine to just `fmt.Fprintf("%#v\n", a.runtime.variables)` and be done with it. Simply using the tried-and-tested fmt package means there's no reason to explicitly test the dump() output formatting in jet.

sauerbraten:
The changes look to me like you addressed #111, not #187?

> To be honest, I am not sure why comparison against master contains also changes implemented by 4beb915 and f67928d - that is confusing.

I assume you didn't pull the newest master of CloudyKit/jet into your fork before starting with the new PR? The hashes are different from your original commits since in this repository, we use a "rebase, then merge" strategy that avoids merge commits. A side effect is that the commits are rewritten with the merging person added to them as commiter (in addition to you as the commit author). See here for example: https://github.com/CloudyKit/jet/commit/f67928da339229e4ccb9b198a6428b8f077cf1c9. This means your fork has a commit with the same changes as f67928d, but it has a different hash and so Git doesn't know the code is the same in your fork and this repo, so the changes are shown in the diff.

Assuming you really did mean to address #111, I added a comment there that is helpful feedback to this PR at the same time: https://github.com/CloudyKit/jet/issues/111#issuecomment-771609451

I don't expect all of that to be done in this PR, but I'd like whatever changes you are able to make to go in that direction, maybe start with just a simple dumpVariables function without any parent scope walking here and we can expand to all the other stuff I laid out in the comment later.

sauerbraten:
As for the failing tests, it seems you created devdump.jet on Windows with \r\n line endings, when Jet only produces \n line endings, so the expected output doesn't match what dump prints: https://ci.appveyor.com/project/CloudyKit/jet/builds/37566590/job/jo1ne5y95yyqs0jd#L880

jan-herout:
Thanks for both comments. When possible, I will look at that. And indeed, this should be related to #111 - my bad.
And your Windows comment is spot on.

jan-herout:
All right. I would appreciate your feedback.  Let's assume that this is the alfa version. Also, documenation is not entirely up to-date, and newly added functions under `dump.go` probably deserve a few well placed comments.

I did not manage to commit the test case template ins uch a manner, that EOL convention would be Unix - sorry for that - however, the test case accounts for that, and runs both on Windows and Linux correctly.

Functions used in the code are "not panicky", I did not like the idea of that.

Regards,

Jan

dump.go
```diff
@@ -0,0 +1,97 @@
...
+//  - everything in Runtime.variables
+//  - everything in Runtime.set.globals
+//  - everything in Runtime.blocks
+func dumpAll(a Arguments, depth int) reflect.Value {
+	var b bytes.Buffer
+	var vars VarMap
+
+	ctx := a.runtime.context
+	fmt.Fprintln(&b, "Runtime.context:")
+	fmt.Fprintf(&b, "\t%s %q\n", ctx.Type(), ctx)
```
sauerbraten, line 22: could we use `%#v` as format of the value we print in all the printf calls? that should make the output more clear: https://play.golang.org/p/6OWfcKwTPmJ

dump.go
```diff
@@ -0,0 +1,97 @@
...
+
+	fmt.Fprintln(&b, "Runtime.variables:")
```
sauerbraten, line 24: can we just call `dumpScope(&b, a.runtime, 0)` here? and then wrap that call in a loop that calls dumpScope with each depth until the requested depth is reached?

dump.go
```diff
@@ -0,0 +1,97 @@
...
+	fmt.Fprintln(&b, "Runtime.blocks:")
+	for _, k := range blockKeys {
+		block := a.runtime.blocks[k]
+		dumpBlock(&b, block)
+	}
+
+	return reflect.ValueOf(b.String())
+}
+
+func dumpScope(w io.Writer, scope *scope, maxDepth, curDepth int) {
```
sauerbraten, line 51: since this is called dumpScope(), I would expect it to dump a single scope. I'd rather call it in a loop than to rely on recursion here.

dump.go
```diff
@@ -0,0 +1,97 @@
...
+	if maxDepth >= curDepth || scope == nil {
+		return
+	}
+	tabs := strings.Repeat("\t", curDepth+1)
+	fmt.Fprintf(w, "%sRuntime.parent.variables, depth=%d\n", tabs, curDepth)
```
sauerbraten, line 56: maybe something like "variables in scope %d levels up" would work better to use this function for the current as well as parent scopes?

eval.go
```diff
@@ -98,6 +99,15 @@ type scope struct {
 	blocks    map[string]*BlockNode
 }
 
+func (s scope) sortedBlocks() []string {
```
sauerbraten, line 102: not sure sorting blocks by name is actually helpful. the most useful way to present them in the dump would be in the order they were defined in, but that's impossible, so just printing in random order is okay IMO.

lex.go
```diff
@@ -287,7 +287,14 @@ func (l *lexer) run() {
 // state functions
 func lexText(l *lexer) stateFn {
 	for {
-		if i := strings.IndexByte(l.input[l.pos:], l.leftDelim[0]); i == -1 {
+		// without breaking the API, this seems like a reasonable workaround to correctly parse comments
```
sauerbraten, line 290: you still haven't rebased your branch on top of the current master of CloudyKit/jet

sauerbraten (changes requested):
I added a few comments on things that jumped out to me on a first look at the changes, there may be more things coming later. It's definitely close to what I imagined; thanks once again!

jan-herout:
Thanks for the review, will look at these changes most likely yesterday.  This is becoming a valuable experience.

dump.go
jan-herout: resolved

dump.go
jan-herout: Resolved. Prepared a new function named `dumpParentScopes`, which contains the cycle. `dumpScope` now prints only current scope.

dump.go
jan-herout: resolved

dump.go
jan-herout: Resolved.

eval.go
jan-herout: I agree that the order in which blocks were created would probably be the most helpful, however disagree with the notion of random ordering. 

Perhaps it is better to get predictable ordering, so that if you make a subtle change to the template, and compare differences between current rendering and previous rendering, you do not have to bother yourself with differences that are not meaningful?

lex.go
jan-herout: resolved, forced push, which might complicate matters.

dump.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// dumpScopeVars prints all variables in the scope.
+func dumpScopeVars(w io.Writer, scope *scope, lvl int) {
+	if scope == nil {
+		return // do not panic if something bad happens
+	}
+	tabs := strings.Repeat("\t", lvl) // delimiter
+	if lvl == 0 {
+		fmt.Fprint(w, "Variables in scope:\n")
```
sauerbraten, line 65: Looking at the test ouput, I think it would be nice for this line to say "Variables in current scope".

dump.go
```diff
@@ -0,0 +1,110 @@
...
+	} else {
+		fmt.Fprintf(w, "%sVariables in scope %d level(s) up:\n", tabs, lvl)
```
sauerbraten, line 67: This line, I'd prefer not to be indented.

dump.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	vars := scope.variables
+	for _, k := range vars.SortedKeys() {
+		fmt.Fprintf(w, "\t%s%s=%#v\n", tabs, k, vars[k])
```
sauerbraten, line 71: Here, let's indent only by one tab, so the different scope levels are listed one after the other, but not nested.

sauerbraten (approved):
I like it! It's immediately useful to users, so I'd say we merge it as soon as the scope indentation is addressed! 👍

eval.go
sauerbraten: Yes, I can see the benefit there. Let's go with that!

dump.go
jan-herout: 4ba85eaabc9dadf230b2778ef6d6c09839d67a28

dump.go
jan-herout: 4ba85eaabc9dadf230b2778ef6d6c09839d67a28

dump.go
jan-herout: 4ba85eaabc9dadf230b2778ef6d6c09839d67a28

jan-herout:
Let me know should you find anything else.
Regards,

Jan

sauerbraten:
Very sorry it took me so long to get back to you!

PR #197, opened 2021-10-27
Put back YieldTemplate

SVyatoslavG:
I put back YieldTemplate which was removed earlier.
I am using it in my code to "insert" template files.
Please put it back.
Thank you very much in advance.


sauerbraten:
How is this different from `t, err := set.GetTemplate("foo")` and then `t.Execute(buffer, nil, context)`? Of course you will have to pass the output writer around outside of the set, but I think the new API is more flexible for rendering single templates.

PR #200, opened 2022-12-08
doc: Expand documentation for Ranger()

matheusd:


sauerbraten:
Thanks a lot! Feel free to complete the markdown docs as well, I saw the section on custom rangers is still marked as TODO: https://github.com/CloudyKit/jet/blob/master/docs/syntax.md#range

sauerbraten:
Oh and if you could describe the gist of the changes in one short sentence I can use in release notes, that would be great! I guess this is not a breaking change, but could some users be affected when they accidentally relied on the buggy behavior?

matheusd:
Will do. For future reference, what is your preferred method to receive the short sentences for use in release notes? In the PR itself? Or maybe we can start tracking such changes in a `docs/next-release-changes.md` (or whatever) file and enforce any relevant PRs/commits to have a line there?

sauerbraten:
I think it's enough if I ask for such a summary in the PRs when I don't think we have a deep enough understanding to write them myself. So feel free to write a comment here for now 👌

matheusd:
Release note:

> Fixed a panic generated when ranging over a channel or other custom rangers without setting the value to variable.

PR #202, opened 2022-12-12
ranger: Improve intsRanger performance

matheusd:
This uses an indirection trick to avoid an allocation in intsRanger.Range call, signficantly improving its performance during iteration.

The trick involves calling reflect.ValueOf on the address of the field instead of using its concrete value. This avoids an allocation performed by the Go runtime to convert the int value to an interface value.

In order to use this trick, the interpretation of the 'i' field needed to be changed from "next index" to "previous index" and its initialization needed to be offset by -1.

Following is the impact of this change in the relevant benchmark:

```
name          old time/op    new time/op    delta
IntsRanger-4    61.0ns ± 2%    23.1ns ± 1%   -62.12%  (p=0.000 n=10+9)

name          old alloc/op   new alloc/op   delta
IntsRanger-4     15.0B ± 0%      0.0B       -100.00%  (p=0.000 n=9+10)

name          old allocs/op  new allocs/op  delta
IntsRanger-4      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

Additionally, it adds a benchmark to assert custom renderer performance.

It's now possible to execute templates with guaranteed zero memory allocations (during execution) by using a combination of int and custom rangers and either simple field access or custom renderers.

### DIFF
--- a/docs/syntax.md
+++ b/docs/syntax.md
@@ -35,6 +35,7 @@
     - [Slices / Arrays](#slices--arrays)
     - [Maps](#maps)
     - [Channels](#channels)
+    - [Custom](#custom-ranger)
     - [else](#else)
   - [try](#try)
   - [try / catch](#try--catch)
@@ -366,7 +367,7 @@ Use `range` to iterate over data, just like you would in Go, or how you would us
         {{.}}
     {{ end }}
 
-Jet provides built-in rangers for Go slices, arrays, maps, and channels. You can add your own by implementing the Ranger interface. TODO
+Jet provides built-in rangers for Go slices, arrays, maps, and channels. You can add your own by implementing the Ranger interface.
 
 #### Slices / Arrays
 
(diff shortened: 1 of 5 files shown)


docs/syntax.md
```diff
@@ -366,7 +366,7 @@ Use `range` to iterate over data, just like you would in Go, or how you would us
         {{.}}
     {{ end }}
 
-Jet provides built-in rangers for Go slices, arrays, maps, and channels. You can add your own by implementing the Ranger interface. TODO
+Jet provides built-in rangers for Go slices, arrays, maps, and channels. You can add your own by implementing the Ranger interface.
```
sauerbraten, line 369: do you intend to send another PR with more explanation about custom rangers? if not, we shouldn't remove this TODO yet, right?

sauerbraten:
Very cool, I think the new intsRanger code is still pretty easy to read, so am definitely gonna merge this. Thank you!

docs/syntax.md
matheusd: [Updated](https://github.com/CloudyKit/jet/compare/a56a8e1c4ae43c8fb5d16aee0cf71df931f79293..2544d9a974cd3b2003195a4b623781c4c510f43b) the docs with a section for the custom ranger, linking to the go doc page for the interface. I've also added a go-styled example (it will show up in the official v6 docs page after this is merged).

PR #213, opened 2024-02-18
Allow DevelopmentMode to be set on or off.

ryanlath:
Allowing DevelopmentMode to be set on or off via a variable is a common pattern I missed.

### DIFF
--- a/examples/asset_packaging/main.go
+++ b/examples/asset_packaging/main.go
@@ -42,7 +42,7 @@ func init() {
 
 	views = jet.NewSet(
 		httpfsLoader,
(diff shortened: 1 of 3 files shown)


sauerbraten:
This makes a lot of sense, thanks for the PR! Two small things:
1. Can you make the existing `InDevelopmentMode` call the new function you added with `true`, so it's clear they are related?
2. I'd rather not mark functions as deprecated without a major version release, which I don't see necessary as yet. Some linters complain about using a function that's marked deprecated, and I don't want users' code to throw a deprecation warning after just a patch or minor release.

Issue #191, opened 2021-03-18
Adding support for Go 1.16's FS interface


advanderveer:
Go 1.16 introduces the fs package and a FS interface. Implementing a Jet loader for this seems like an ideal use case. Are there any plans for this, would it be easy?

sauerbraten:
Yes, that adapter should be all you need to use 1.16's FS interface!
Let me know if it works, It probably makes sense to have this boilerplate directly in Jet.

advanderveer:
I actually just found a way to do this without an extra type by using jets `loaders/httpfs` package:

```
//go:embed *.html
var tfs embed.FS

// loader allows for loading templates
func loader() (jet.Loader, error) {
	return httpfs.NewLoader(http.FS(tfs))
}
```

it is a bit weird to require a "httpfs/http" package for something that just have to deal with filesystems but it works as expected.

replay111:
@advanderveer How did you pass this func loader to Jet Engine so it is loading files from ```tfs``` ?

replay111:
Hello again,

I've tried this way:

```
var view *jet.Set

func init() {
	httpFsLoader, err := loader()
	if err != nil {
		log.Panicln("error: ", err)
	}
	view = jet.NewSet(
		httpFsLoader,
		jet.InDevelopmentMode(),
	)
}

func loader() (jet.Loader, error) {
	return httpfs.NewLoader(http.FS(Repo.App.TemplatesFS))
}
```

where Repo.App.TemplatesFS stores:

```
type ApplicationConfig struct {
	InfoLog       *log.Logger
	ErrorLog      *log.Logger
	InDevelopment bool
	Session       *scs.SessionManager
	TemplatesFS   embed.FS
}

//go:embed templates
var assets embed.FS
var app config.ApplicationConfig

app.TemplatesFS = assets

```

But compiler does not like this part:

```
func loader() (jet.Loader, error) {
	return httpfs.NewLoader(http.FS(Repo.App.TemplatesFS))
}
```
error:
cannot use httpfs.NewLoader(http.FS(Repo.App.TemplatesFS)) (value of type jet.Loader) as jet.Loader value in return statement: wrong type for method Open (have func(templatePath string) (io.ReadCloser, error), want func(templatePath string) (io.ReadCloser, error))compilerInvalidIfaceAssign

but Repo.App.TemplatesFS is embed.FS so it is the same type as in example from @advanderveer

replay111:
Ok I also found solution for my case:

```
[...]

PR #216, opened 2024-03-03
Add go1.16 embed.FS loader

ryanlath:
Adds embed.FS Loader.  Requires go version 1.16
Addresses: https://github.com/CloudyKit/jet/issues/191


loaders/embedfs/loader.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Open implements Loader.Open() on top of an embed.FS.
+func (l *embedFileSystemLoader) Open(name string) (io.ReadCloser, error) {
+	return l.fs.Open(filepath.Join(l.dir, filepath.FromSlash(name)))
+}
+
+// Exists implements Loader.Exists() on top of an embed.FS by trying to open the file.
+func (l *embedFileSystemLoader) Exists(name string) bool {
+	_, err := l.fs.Open(filepath.Join(l.dir, filepath.FromSlash(name)))
+	return err == nil && !os.IsNotExist(err)
```
sauerbraten, line 33: Since embed.FS.Open() might return a directory "file", we should also check for that using [fs.File.Stat()](https://pkg.go.dev/io/fs#File) and [IsDir](https://pkg.go.dev/io/fs#FileInfo). See https://github.com/CloudyKit/jet/blob/1926e5e37df90038229356a265a22d7d609b50ff/loader.go#L74 where something similar is done.

ryanlath: done.

PR #223, opened 2025-03-23
Update link for Go API to point to pkg.go.dev instead of beta.pkg.go.dev

jhard:
beta.pkg.go.dev requires a login


sauerbraten:
good catch, thank you!

